PR #6, opened 2013-04-29
Claudiu gcc4.8 delay slot sched fix

claziss:


simonpcook:
These commits have been squashed and added to the 4.8 tree.

PR #11, opened 2013-06-10
Add alias:-mEM, -mARCv2EM. Add asl insn.  No barrel_shifter for EM (default)

claziss:


amylaar:
Quoting Claudiu Zissulescu notifications@github.com:

> - add ldh_s.x instruction
>   :+  [(set (match_operand:SI 0 "dest_reg_operand" "=Rcqq,w,Rcq,r")
>   that constraint should really be:                         Rcqq
>   so that you can drive register allocation for -mmixed-code.
> - Initial work in enabling ARC EM specific insns
> - Merge branch &#39;arc_4_8-gcc-4_8-dev&#39; of  
>   github.com:foss-for-synopsys-dwc-arc-processors/gcc into  
>   arc_4_8-gcc-4_8-dev
> - Add mEM, mARCv2EM alias; Remove force_short; Add asl a,b insn  
>   for EM; Barrel shifter is not default for EM (Fix)
>   :+2013-06-10 Claudiu Zissulescu claziss@synopsys.com
>   :+
>   :+  \* config/arc/arc.opt: Remove force_short option
>   :+  \* config/arc/arc.h: Likewise

OK.

:+  \* common/config/arc/arc-common.c: Likewise; EM default has no
:+        barrel shifter.

If that is true, you should match the ARC601 multilibs when the EM has not
its own multilibs.  Vice versa, when you are not building the ARC700
multilibs, I think the best match for ARC700 is ARC600 -mnorm.

:+        \* config/arc/arc.md: Add asl instruction for EM

What is that for?  Just to be able to use asl_s - which is not EM specific
to start with?  asl is slower than add, so it shouldn't be used when
compiling for speed.  If you really think it's important when compiling
for size (why, if that is important, not have the asl multiple insn?),
you should make shift_si3 into a splitter, and for the one-bit shift left,
the enabled attribute of the short alternative should be depend on
(match_test "optimize_function_for_size_p (cfun)") .

:+  \* config/arc/arc.opt: Add alias for mEM and mARCv2EM

Don't do that.  We want DRIVER_SELF_SPECS to canonicalize the  
subtarget selection to -mcpu= options, so that the other specs have  
just that
option to check.

I also find the git merging makes it very hard to tell what patch is actually
up for review.

PR #14, opened 2013-06-19
Fix passing mswap, mnorm, and mbarrel-shifter options to gas from gcc driver (EM case)

claziss:


jeremybennett:
Claudiu,

I've asked Joern to review this as his first priority tomorrow, Thursday.

claziss:
I will make the necessary modification in GAS. Hence, having an coherent gas interface.

PR #29, opened 2013-11-07
libstdc++: Disable gc-sections to work around binutils bug

anthony-kolesov:
I don't think this should go into mainline, as this is a workaround usefull only for released source code.

This is to work around the following error message:
BFD (GNU Binutils) 2.23.2 assertion fail elf32-arc.c:2140

Signed-off-by: Mischa Jonker mjonker@synopsys.com
Signed-off-by: Anton Kolesov anton.kolesov@synopsys.com


jeremybennett:
I've been discussing this with Joern. He wants to know more about the assertion failure, so perhaps you could join him on IRC (#arcfoss) to discuss.

anthony-kolesov:
Closing this, because fix to binutils has been applied instead of this one.

PR #32, opened 2014-11-24
Reduce the cases where TLS fixups are changed to LE.

T-J-Teru:
Change TLS fixup type to Local Exec in a reduced numbed of cases.  Add some additional support for handling GD and IE fixups.

You'll need gdb-binutils including this patch: https://github.com/foss-for-synopsys-dwc-arc-processors/binutils-gdb/pull/3

### DIFF
--- a/gcc/ChangeLog.ARC
+++ b/gcc/ChangeLog.ARC
@@ -1,3 +1,12 @@
+2014-11-18  Andrew Burgess  <andrew.burgess@embecosm.com>
+
+        * config/arc/arc.c (arc_print_operand): Handle global dynamic,
+	UNSPEC_TLS_GD, fixups.
+	(arc_print_operand_address): Handle Imediate Exec, UNSPEC_TLS_IE,
+	fixups.
+	(arc_legitimize_tls_address): Change Local Dynamic to Local Exec
+	when we're not compiling PIC code.
+
 2013-07-29 Claudiu Zissulescu <claziss@synopsys.com>
 
 	* config/arc/arc.c (arc_setup_incoming_varargs): Pass
--- a/gcc/config/arc/arc.c
+++ b/gcc/config/arc/arc.c
@@ -3502,11 +3502,13 @@ arc_print_operand (FILE *file, rtx x, int code)
       if (flag_pic
 	  || (GET_CODE (x) == CONST
 	      && GET_CODE (XEXP (x, 0)) == UNSPEC
-	      && XINT (XEXP (x, 0), 1) == UNSPEC_TLS_OFF)
+	      && (XINT (XEXP (x, 0), 1) == UNSPEC_TLS_OFF
+	          || XINT (XEXP (x, 0), 1) == UNSPEC_TLS_GD))
 	  || (GET_CODE (x) == CONST
 	      && GET_CODE (XEXP (x, 0)) == PLUS
 	      && GET_CODE (XEXP (XEXP (x, 0), 0)) == UNSPEC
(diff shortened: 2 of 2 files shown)


amylaar:
ChangeLog typo: arc_print_operand_address

The change to arc_legitimize_tls_address is nonsentical.

T-J-Teru:
I've forced the commit on my branch to fixup the typos in the changelog.  Nothing has changed in the code.

Joern, please could you expand on "nonsentical", even after looking that word up in a dictionary I am no closer to understanding your second sentence.

amylaar:
On 4 December 2014 at 09:24, Andrew notifications@github.com wrote:

> I've forced the commit on my branch to fixup the typos in the changelog.
> Nothing has changed in the code.
> 
> Joern, please could you expand on "nonsentical", even after looking that
> word up in a dictionary I am no closer to understanding your second
> sentence.
> 
> I meant nonsensical.
> You are not just reducing the cases where the tls model is forced to LE,
> you are also expanding them.  To the point of making the code worse in some
> cases, ignoring the requests by the programmer.

simply reducing would be:

if (!flag_pic && model == TLS_MODEL_LOCAL_DYNAMIC && TARGET_TLS9)

cases that could be legitimately added (back):
- GD where the referenced symbol is not declared extern (i.e. it'll
  end up being defined in the executable)
- LD where there is no base symbol, or were the base symbol has no
  tls9 attribute.

T-J-Teru:
I don't understand why the size of the reference has any relevance to the style of access that is required?  The size surely comes into play when generating code for the access; what code to generate depends on the size.  Maybe you could give an example that concerns you and we/I could investigate the code size implications.

I think that the conversion of GD references is the actual cause of the bug that we originally saw.  The solution I have put forward seemed to me, to be more inline with other targets.  The optimisation you propose does not feel ARC specific, could you point out another target that does the same optimisation at this place in the code, then I'll be able to see how it should be done.

amylaar:
On 4 December 2014 at 10:07, Andrew notifications@github.com wrote:

> I don't understand why the size of the reference has any relevance to the
> style of access that is required?

The only reason to change LD to LE is to increase code density and/or speed.
If the LD reference is both shorter and faster than the LE reference, we
shouldn't change it to LE.

> The size surely comes into play when generating code for the access; what
> code to generate depends on the size. Maybe you could give an example that
> concerns you and we/I could investigate the code size implications.

 **thread int base = 0;
__thread int a __attribute**((tls_model("local-dynamic"),tls9("base"))) = 0;
**thread int b __attribute**((tls_model("local-dynamic"),tls9("base"))) = 0;
**thread int c __attribute**((tls_model("local-dynamic"),tls9("base"))) = 0;
**thread int d __attribute**((tls_model("local-dynamic"),tls9("base"))) = 0;

int f ()
{
  return a+b+c+d;
}

I think that the conversion of GD references is the actual cause of the bug

> that we originally saw. The solution I have put forward seemed to me, to be
> more inline with other targets. The optimisation you propose does not feel
> ARC specific, could you point out another target that does the same
> optimisation at this place in the code, then I'll be able to see how it
> should be done.

All I'm asking you is to refrain from pessimizing the code - and addressing
modes are always target specific, and hence we got target-specific cost
considerations.

T-J-Teru:
Have discussed with @amylaar  and we agree that in the review time we have had available we can find no new functional regressions introduced by this change.  Given that the patch improves TLS functionality in known cases, and causes no regressions in the GCC and uClibc testsuites, we agree that the patch should be pushed.

amylaar:
I agree.

PR #36, opened 2015-04-13
Restore blink register after millicode optimisation.

T-J-Teru:
If the blink register is read in a function that has made use of the
millicode optimisation to spill registers to the stack then, previously,
we would read an invalid value for the blink register.  This patch adds
code to restore the blink register from the stack, ensuring that the
correct value of blink is read.

gcc/ChangeLog:

```
* config/arc/arc.c (arc_expand_prologue): Restore blink register
if needed after millicode optimisation.
```

### DIFF
--- a/gcc/ChangeLog.ARC
+++ b/gcc/ChangeLog.ARC
@@ -1,3 +1,8 @@
+2015-04-08  Andrew Burgess  <andrew.burgess@embecosm.com>
+
+	* config/arc/arc.c (arc_expand_prologue): Restore blink register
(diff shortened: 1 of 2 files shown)


claziss:
We need to instruct the compiler that millicode routine is clobbering the return address. Unfortunately, the proposed solution is not it. I will close this pull request with the hope we will get the right solution.

T-J-Teru:
As discussed via email,  I think we both agree that this patch is probably both safe, and the simplest way to resolve the `blink` clobbering issue that we're seeing.

PR #39, opened 2015-06-03
Add -mno-sdata option to the test's compilation options

kolerov:
SDATA section is forced by default and ARC compiler decides
to omit global constants references before linking stage.
Thus this test fails:
`gcc/testsuite/gcc.dg/tree-ssa/ssa-store-ccp-2.c`

It's necessary to pass -mno-sdata option to the compiler
to not use SDATA section.

STAR: 9000905889

Signed-off-by: Yuriy Kolerov yuriy.kolerov@synopsys.com

### DIFF
--- a/gcc/testsuite/gcc.dg/tree-ssa/ssa-store-ccp-2.c
+++ b/gcc/testsuite/gcc.dg/tree-ssa/ssa-store-ccp-2.c
@@ -1,5 +1,6 @@
 /* { dg-do compile } */
 /* { dg-options "-O2 -fdump-tree-optimized" } */
+/* { dg-options "-O2 -fdump-tree-optimized -mno-sdata" { target arc*-*-elf32 } } */
 
 const int conststaticvariable;
 
--- a/gcc/ChangeLog.ARC
+++ b/gcc/ChangeLog.ARC
@@ -1,3 +1,8 @@
+2015-06-03  Yuriy Kolerov  <yuriy.kolerov@synopsys.com>
+
+	* gcc/testsuite/gcc.dg/tree-ssa/ssa-store-ccp-2.c: Add
+	-mno-sdata option to the test's compilation options.
+
 2015-05-26  Claudiu Zissulescu <claziss@synopsys.com>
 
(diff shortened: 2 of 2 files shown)


gcc/testsuite/gcc.dg/tree-ssa/ssa-store-ccp-2.c
```diff
@@ -1,5 +1,5 @@
 /* { dg-do compile } */
-/* { dg-options "-O2 -fdump-tree-optimized" } */
+/* { dg-options "-O2 -fdump-tree-optimized -mno-sdata" } */
```
anthony-kolesov: Probably `-mno-sdata` should be arch dependent - appended only for ARC. Is that possible?

kolerov: Yes, it's possible by adding this line:
`/* { dg-options "-O2 -fdump-tree-optimized -mno-sdata" { target arc*-*-* } } */`
I'll fix it.

anthony-kolesov: -O2 and dump tree should be not arch dependent. So like this?
```
/* { dg-options "-O2 -fdump-tree-optimized" { target arc*-*-* } } */
/* { dg-options "-O2 -fdump-tree-optimized -mno-sdata" { target arc*-*-* } } */
```
I don't really know.

gcc/testsuite/gcc.dg/tree-ssa/ssa-store-ccp-2.c
```diff
@@ -1,5 +1,6 @@
 /* { dg-do compile } */
 /* { dg-options "-O2 -fdump-tree-optimized" } */
+/* { dg-options "-O2 -fdump-tree-optimized -mno-sdata" { target arc*-*-* } } */
```
claziss: The triple should only select the ELF toolchain as for uClibc we do not use sdata.

kolerov:
I've fixed the triple for ARC.

PR #40, opened 2015-06-04
Mark some -fcompare-debug tests as expected fails

kolerov:
These tests always fail because of the feature of the current ZOL implementation for ARC targets.

Point is that one value in the internal representation always differs for debug and not debug configuration. The value that is changing is the INSN_UID of loop end insn. The INSN_UID may vary from compilation to compilation. Hence, the error.

When ZOL implementation will be changed these tests must be reviewed.

STAR: 9000906644

Signed-off-by: Yuriy Kolerov yuriy.kolerov@synopsys.com

### DIFF
--- a/gcc/ChangeLog.ARC
+++ b/gcc/ChangeLog.ARC
@@ -1,3 +1,19 @@
+2015-06-03  Yuriy Kolerov  <yuriy.kolerov@synopsys.com>
+
+	* gcc/testsuite/c-c++-common/pr44832.c: Mark as expected
+	fail for ARC targets because -fcompare-debug check always
+	fails because of the feature of the current ZOL
+	implementation.
+	* gcc/testsuite/gcc.dg/pr42630.c: Likewise.
+	* gcc/testsuite/gcc.dg/pr42631.c: Likewise.
+	* gcc/testsuite/gcc.dg/pr42916.c: Likewise.
+	* gcc/testsuite/gcc.dg/pr44023.c: Likewise.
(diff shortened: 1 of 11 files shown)


jeremybennett:
Yuriy.  This is not a correct use of XFAIL - this should be used for tests that are inherently going to fail because of some intrinsic property of the system.  A test that is XFAIL will always be XFAIL.

In this case, the failure is because of a feature of the compiler, which could be fixed in the future.  I'm not sure if GCC allows KFAIL (this may still only be possible in GDB), which indicates a failure in the current implementation which is known, and may be fixed in the future.  Otherwise the test should remain as PASS - that is its expected behavior with the ultimate perfect compiler.

HTH

anthony-kolesov:
Hi Jeremy,

Yuriy implements my request in this regard. 

There is an intrinsic problem problem of current ZOL implementation in GCC, which causes this "failure". What I want is that FAIL in dejagnu result should represent real failure which will be fixed. The ZOL issue hasn't been fixed in 20 months by now, and it is not clear if it will be fixed in next 20 months or 200. So from my perspective FAIL is fail, and if that is fail - it must be fixed. If GCC developers declare that this FAIL is not a fail, and they are not going to fix that in a realistic timeframe (until next release) - then they need to take an action that this test will not be FAIL, but would be something else. So in theory our customer will take tools and will run regression and will get 0 fails. With current situation they will get non-zero number of failures and there will be questions.

anthony-kolesov:
Closing PR after negative feedback from everyone.

PR #42, opened 2015-08-31
GCC generating out of range branch instructions.

T-J-Teru:
This series addresses a couple of issues where either instructions are under reporting their length, or where extra nop instructions are being added to the instruction stream at a very late stage.  Both of these problems cause branch shortening to make an incorrect choice about when a short branch can be used.  This in turn causes the generated assembler to fail to assemble.

This series starts with a few small clean up patches, the last two patches in the series are the real work.  Please see the individual commit messages for more detail on each patch.


claziss:
Thank u, Andrew

PR #43, opened 2015-09-08
gcc: Fix generated assembler for doloop_begin_i.

T-J-Teru:
Commit 23a04c4 removed some duplicate code from doloop_begin_i, however,
the second version of the code, which was never reached before, but now
is, was not quite correct; it failed to make use of the instruction
operands, instead hard coding values into the generated assembler.

Now that commit 23a04c4 has been made, the previously unused, second
piece of code is being tested, this commit fixes the problems by
updating the generated assembler to match the original that was removed
in 23a04c4, this should resolve all issues.

gcc/ChangeLog:

```
* config/arc/arc.md (doloop_begin_i): Make correct use of
parameters in generated assembler.
```

### DIFF
--- a/gcc/ChangeLog.ARC
+++ b/gcc/ChangeLog.ARC
@@ -1,3 +1,8 @@
+2015-09-08  Andrew Burgess  <andrew.burgess@embecosm.com>
+
+	* config/arc/arc.md (doloop_begin_i): Make correct use of
+	parameters in generated assembler.
+
 2015-08-29  Andrew Burgess  <andrew.burgess@embecosm.com>
 
 	* config/arc/arc.md (pass_arc_hazard_avoidance): New variable.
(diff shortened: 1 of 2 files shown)


claziss:
Ah... magic numbers.It looks good. Thank you Andrew

anthony-kolesov:
Hi guys,

This commit introduces an error when building HS uClibc toolchain. See attached test case
[cp-demangle.i](https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/files/3791/cp-demangle.i)

```
arc-linux-gcc -O2 -fPIC cp-demangle.i -c -marchs
/tmp/cchqY01s.s: Assembler messages:
/tmp/cchqY01s.s:6038: Error: junk at end of line: `@pcl'
```

That should have been found quicker by the nightly build, however unfortunately there is an another issue in my verification scripts, that masked this one. Both -O2 and -fPIC are required to reproduce.

T-J-Teru:
@anthony-kolesov I believe I have a fix for this issue in https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/45

PR #44, opened 2015-09-25
Add ARC naked function support.

T-J-Teru:
This series of commits adds support for naked functions.  The real work is in commit d9ed07c, all of the other commits are clean up work in preparation for naked function support.


T-J-Teru:
I just found a regression caused by this pull request.  Please don't pull until I've updated.

claziss:
Thanks Andrew!

PR #45, opened 2015-09-29
arc: doloop_begin_i, handle NULL_RTX case.

T-J-Teru:
In some cases the loop_start label can be NULL_RTX, this is when the
start of the loop immediately follows the doloop_being_i instruction, in
these cases we don't have a code label for the loop start, and so should
not try to generate a '@symbol@pcl' style relocation.  Instead, we
should hard-code the relative location of the start of loop instruction
into the instruction stream.

This commit continues the fix started in commit 20fc569, which
completely removed the hard coded immediate as a result of hitting a
case where the loop_start was a label that did not work with a hard
coded immediate.

gcc/ChangeLog:

```
* config/arc/arc.md (doloop_begin_i): Handle case where
loop_start is NULL_RTX.
```

### DIFF
--- a/gcc/ChangeLog.ARC
+++ b/gcc/ChangeLog.ARC
@@ -1,3 +1,8 @@
+2015-09-29  Andrew Burgess  <andrew.burgess@embecosm.com>
+
+	* config/arc/arc.md (doloop_begin_i): Handle case where
+	loop_start is NULL_RTX.
+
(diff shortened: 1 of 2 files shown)


anthony-kolesov:
I'm merging this PR, since Claudiu allowed me, but he might want to review the code change more thoroughly tomorrow. For now I just want nightly builds to be fixed.

PR #61, opened 2017-01-19
[ARC] add support for enter_s/leave_s instructions

calandoa:
This patch implement the use of enter_s/leave_s opcodes.
It is enabled by -mcode-density-frame.

Cases not implemented yet are: IRQ functions, frame pointer used, exception handler used, non ARCv2 

The only testing done has been manual inspection of the objdump and build/run on two different projects (no idea on how to run automated test suites).

Size change of the text section:
Project 1 (-Os): 108408
                          104072   (->96%)

Project 2 (-Os): 187252
                          175240 (->93.6%)

Project 2 (-O2): 245936
                          219584 (-> 89.3%)

The code in the .md files may be a bit poor. First time I use this syntax and not sure how to cope with parallel rtx.

### DIFF
--- a/gcc/config/arc/arc.c
+++ b/gcc/config/arc/arc.c
@@ -2703,6 +2703,7 @@ arc_compute_frame_size (void)
   /* Saving blink reg in case of leaf function for millicode thunk calls.  */
   if (optimize_size
       && !TARGET_NO_MILLICODE_THUNK_SET
+      && !TARGET_CODE_DENSITY_FRAME
       && !crtl->calls_eh_return)
     {
       if (arc_compute_millicode_save_restore_regs (gmask, frame_info))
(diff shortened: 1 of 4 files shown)


claziss:
Hi,

I am back in the office, and I will give you an feedback by the end of next week.

Thank you for your interest in our tools

claziss:
The enter/leave functions are not ABI compatible thus care needs to be taken when using them. 
Here are a number of observations:
- It seems to me the dwarf info is wrong.
- It also looks to me the original millicode thunk predicates can be adapted to be used with enter/leave.
- I would really want to see this work integrated with millicode thunk implementation.
- You do not say how did you test it. I would suggest to run gcc's dejagnu tests with this new option enabled.
- For size measurements, we use CSiBE benchmark. 

Thanks,

calandoa:
Hi,

Just out of curiosity, what is the difference with the ABI ? The only thing I would think of is the regsiter saving order in the frame, but it looks the same to me.

Regarding the patch:
- dwarf info: no idea, I did not check this at all as I do not use it.
- re-using thunk predicates: not sure here, as the epilogue is really different. Also, I focused on writing a patch the more independant from the current code to be sure not to broke something already working. Even without my patch, the current code could really use some cleanup, but I would postpone this later.
- testing: as said, I did only basic testing with working projects: I checked manually the assembler after building, and run basic tests on these projects. I did not check how to use dejagnu for gcc arc and I suppose specific HW is required.

claziss:
Our abi states the blink goes first on the stack. However, the enter/leave are placing it on the top of it. This should not be a problem for simple programs. A way to go around is to push/pop first blink and then to use enter/leave. Or just leave it as it is but force it off for linux types of systems.

Thanks

claziss:
Hi,
Will you consider to work out a new patch which includes my comments or shall I make it?

Thanks

calandoa:
Hi Claudiu,

Sorry I do not plan to improve this patch. I have no clue on how we could merge the predicate part with the existing one, and also I have no dejagnu setup to test it more toroughly. Of course feel free to reuse it any way you wish, and/or close this issue.

claziss:
Hi,
We will do it, and it will be in the next release. I'll close this thread when it will be ready :)

claziss:
Closing this pull request. The enter/leave instruction support is now in arc-2018.09 (upcoming) release.

PR #83, opened 2019-12-24
[ARC] LIBGCC: Allow configure for olde good 32-bit ARC

abrodkin:
Otherwise libgcc configuration fails like that:
----------------------------->8-------------------------
[ALL  ]    *** Configuration arc-unknown-elf not supported
----------------------------->8-------------------------

Signed-off-by: Alexey Brodkin <abrodkin@synopsys.com>

### DIFF
--- a/libgcc/config.host
+++ b/libgcc/config.host
@@ -95,7 +95,7 @@ amdgcn*-*-*)
 	cpu_type=gcn
 	tmake_file="${tmake_file} t-softfp-sfdf t-softfp"
 	;;
-arc[be]*-*-*)
+arc-*-* | arceb-*-*)
 	cpu_type=arc
 	;;
 arc64-*-*)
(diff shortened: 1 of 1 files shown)


libgcc/config.host
```diff
@@ -407,12 +407,12 @@ amdgcn*-*-amdhsa)
...
 	extra_parts="crt0.o"
 	;;
-arc[eb]*-*-elf*)
+arc-*-elf* | arceb-*-elf*)
 	tmake_file="arc/t-arc"
 	extra_parts="crti.o crtn.o crtend.o crtbegin.o crtendS.o crtbeginS.o"
 	extra_parts="$extra_parts crttls.o"
 	;;
-arc[eb]*-*-linux*)
+arc*-*-linux* | arceb-*-linux*)
```
claziss: maybe it should be: `arc-*-linux*`

abrodkin: Right, spello :) Will fix and push again.

PR #84, opened 2020-01-11
Fix MUL64 registers assignment

petrokarashchenko:


petrokarashchenko:
@claziss please take a look. This intends to fix https://github.com/foss-for-synopsys-dwc-arc-processors/toolchain/issues/234. At least I get correct code generated.
I have checked binutils and see next code there:
```
  declare_register ("mlo", 57);
  declare_register ("mmid", 58);
  declare_register ("mhi", 59);
```
Taking into account this and what I have read in about ARC MUL64 I assume that my changes are correct.

Please review.

petrokarashchenko:
No. Its bad change. It breaks other logic. Need to evaluate a better one

gcc/config/arc/arc.md
```diff
@@ -2271,7 +2271,7 @@ archs4x, archs4xd"
       && !satisfies_constraint_L (operands[1]))
     {
       /* MUL64 <0,>b,s12 00101bbb10000100 0BBBssssssSSSSSS  */
-      int n = true_regnum (operands[0]);
+      int n = DBX_REGISTER_NUMBER(true_regnum (operands[0]));
```
petrokarashchenko, line 2274: @claziss how do you think, do we need to apply `DBX_REGISTER_NUMBER` to all other places where `true_regnum` is used?

claziss:
Thank you for bringing this issue up. GCC implementation for mul64 assumes a co-existence configuration (i.e. r58 is mlo and r59 is mhi). As far as I can understand your ARC600 implementation is having just a mul64 configuration which implies r57 mlo, r58 mmid, and r59 mhi. 
Your patch is not complete as it drops/changes the default gcc support for arc600 potentially leading to tools regression for others. I will take action and cherry pick your pull request in a proper patch for the next 2020.03 release.

petrokarashchenko:
@claziss Thank you very much for review. I do not fully understand what do you mean by "default gcc support for arc600" and "GCC implementation for mul64 assumes a co-existence configuration (i.e. r58 is mlo and r59 is mhi)", but I thank you very much for contributing fix into 2020.03

I thought my changes are safe because of assumption that `mulsi_600` is applied only for ARC600 and should not break other configurations.

Taking into account your response I do not see any option to use `-mmul64` option for my configuration because this will produce "hidden bomb" randomly in my code.

Looking forward to try `2020.03` release.

petrokarashchenko:
@claziss please close this PR as soon as you will submit fix into `2020.03` branch. Just want to keep this open for now not to loose this changes.

claziss:
ARC600 is a configurable template which accept many other extensions. It looks to me the ARC600 support in gcc is quite limited when talking about supporting those ARC600 extensions. It is good you rise the problems about it, and I will try to add the missing features and patching the code asap. However, we need to take care also not to break whatever is supposedly there :) and this requires some extensive testing and test cases created.

claziss:
@petrokarashchenko I am reading my ARC600 documentation and it looks like mlo/mmid/mhi registers are always fixed to same one regardless of the big/little endian CPU configuration. Please can you confirm, namely mlo is at r57?

petrokarashchenko:
@claziss Yes :) I confirm this. Actually this was my biggest confusion, but this `r58`/`r59` and `DBX_REGISTER_NUMBER` mapping. The `mlo` is `r57`, `mmid` is `r58` and `mhi` is `r59`.

petrokarashchenko:
```
.extCoreRegister mlo, 57, r|w
.extCoreRegister mmid, 58, r|w
.extCoreRegister mhi, 59, r|w
.extCoreRegister mlo, 57, r
.extCoreRegister mmid, 58, r
.extCoreRegister mhi, 59, r
```

petrokarashchenko:
@claziss actually I thought that this BIG_ENDIAN trick is done according to some GCC "internal kitchen".
I have read description in `arc.c`
```
/* Core registers 56..59 are used for multiply extension options.
   The dsp option uses r56 and r57, these are then named acc1 and acc2.
   acc1 is the highpart, and acc2 the lowpart, so which register gets which
   number depends on endianness.
   The mul64 multiplier options use r57 for mlo, r58 for mmid and r59 for mhi.
   Because mlo / mhi form a 64 bit value, we use different gcc internal
   register numbers to make them form a register pair as the gcc internals
   know it.  mmid gets number 57, if still available, and mlo / mhi get
   number 58 and 59, depending on endianness.  We use DBX_REGISTER_NUMBER
   to map this back.  */
  char rname56[5] = "r56";
  char rname57[5] = "r57";
  char rname58[5] = "r58";
  char rname59[5] = "r59";
  char rname29[7] = "ilink1";
  char rname30[7] = "ilink2";
```
and was expecting that usage of `DBX_REGISTER_NUMBER` for `int n = DBX_REGISTER_NUMBER(true_regnum (operands[0]));` should fix the issue.

claziss:
I cannot understand what is your issue? Please can you just tell me that :)

petrokarashchenko:
The issue is https://github.com/foss-for-synopsys-dwc-arc-processors/toolchain/issues/234

petrokarashchenko:
Code for `mul64 0, mlo, 1000` should be `2984 7A0F`, but not `2B84 7A0F`.
`2B84 7A0F` use `r59` instead of `r57` (`mlo`).

claziss:
Ok, I missed that entry :) I'll close this thread, and we will continue on that one :)

PR #86, opened 2020-05-10
[Question] How to send change request to GCC mainline?

petrokarashchenko:
[FIX] Correct check of object bounds during address adjustment

[FIX] Propagate uncached type attributes to unaligned types

[ARC] Update tests

gcc/
xxxx-xx-xx  Petro Karashchenko  <petro.karashchenko@ringteam.com>

	* emit-rtl.c (adjust_address_1): Use size of record or
        union instead of memory size attribute for record or
        union type expressions.

testsuite/
xxxx-xx-xx  Petro Karashchenko  <petro.karashchenko@ringteam.com>

        * gcc.target/arc/uncached-9.c: New file.

### DIFF
--- a/gcc/emit-rtl.c
+++ b/gcc/emit-rtl.c
@@ -2356,6 +2356,7 @@ adjust_address_1 (rtx memref, machine_mode mode, poly_int64 offset,
   scalar_int_mode address_mode;
   struct mem_attrs attrs (*get_mem_attrs (memref)), *defattrs;
   unsigned HOST_WIDE_INT max_align;
+  poly_int64 object_size;
 #ifdef POINTERS_EXTEND_UNSIGNED
   scalar_int_mode pointer_mode
     = targetm.addr_space.pointer_mode (attrs.addrspace);
@@ -2456,8 +2457,13 @@ adjust_address_1 (rtx memref, machine_mode mode, poly_int64 offset,
 
   if (maybe_ne (size, 0))
     {
+      /* Use size of record or union type to check object bounds correctly */
+      if (attrs.expr && RECORD_OR_UNION_TYPE_P (TREE_TYPE (attrs.expr)))
+        object_size = int_size_in_bytes (TREE_TYPE (attrs.expr));
+      else
+        object_size = attrs.size;
       /* Drop the object if the new right end is not within its bounds.  */
-      if (adjust_object && maybe_gt (offset + size, attrs.size))
+      if (adjust_object && maybe_gt (offset + size, object_size))
 	{
 	  attrs.expr = NULL_TREE;
 	  attrs.alias = 0;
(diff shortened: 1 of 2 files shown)


petrokarashchenko:
@claziss I have revised my previous changes and made a new improved version.
I truly believe that this should be a correct change and I want to ask on how to submit patches to GCC mainline? Maybe you can guide me a bit or help and try to discuss this change with GCC maintainers. Is Bugzilla issue mandatory for submission of any change?

gcc/emit-rtl.c
```diff
@@ -2456,8 +2457,13 @@ adjust_address_1 (rtx memref, machine_mode mode, poly_int64 offset,
 
   if (maybe_ne (size, 0))
     {
+      /* Use size of record or union type to check object bounds correctly */
```
petrokarashchenko, line 2460: However this does not cover case of flexible array members. Probably we can't make object boundary check in case of structures or records.
```
if (adjust_object && maybe_gt (offset + size, attrs.size)
    && !(attrs.expr && RECORD_OR_UNION_TYPE_P (TREE_TYPE (attrs.expr))))
```

petrokarashchenko:
@claziss 
Actually it is even worse. I can make a test case with
```
typedef volatile struct {
    uint16_t a0[2];
    uint32_t b0[2];
} __attribute__((packed, uncached)) t1_t;

t1_t *p;
```
and depending on access to `p->b[0]`, `p->b[10]`, `p->b[-1]` or `p->b[-10]` I'm getting different results compared to same access done without packing. Without packing I'm getting `.di` instructions in all cases, but with packing I see that I'm actually hitting one of `if (adjust_object && ...` conditions. So the solution can be to add `if (adjust_object && !(attrs.expr && RECORD_OR_UNION_TYPE_P (TREE_TYPE (attrs.expr))) adjust_object = 0`, but `adjust_address_1` with `adjust_object` set to `1` are called only with `adjust_bitfield_address`, `adjust_bitfield_address_size` and `adjust_bitfield_address_nv`, so I assume those to be valid only for records and unions. So then I'm having a question on why do we need to drop expression for bit field adjustments at all?

petrokarashchenko:
I will make few more experiments with pure bit fields (I mean not with unaligned or packed data that are treated the same a bit fields).

claziss:
If you want to submit a patch to mainline, or ARC backend, you just need to send a email to [gcc patches](gcc-patches@gcc.gnu.org) mailing list with an explanation of what you try to fix, and  how u did test it. More [here](https://gcc.gnu.org/contribute.html)

petrokarashchenko:
Currently my understanding and knowledge is not not enough to propose the solution, but what I see is that expression (and attributes) is draped for all members of `packed` type except the first one by one of `adjust_bitfield_address_...` calls.

PR #90, opened 2020-09-01
Shahab review atomic option 2

shahab-vahedi:
This is the sum of both changes so far.


vineetgarc:
Hi folks, can we enable baseline atomics by default - it is unlikely that linux users can do anythign useful w/o LLOCK/SCOND. We can leave the new AMOs as dependent on user provided toggle.

PR #96, opened 2021-06-16
m64/m128: Add cpymem expand

shahab-vahedi:
Use lddl/stdl for moves if -m128 and -munaligned-access is used.
If there is no -m128, but still unaligned access is allowed, then
use the ldl/stl.

### DIFF
--- a/gcc/config/arc64/arc64.c
+++ b/gcc/config/arc64/arc64.c
@@ -3033,6 +3033,35 @@ arc64_move_pointer (rtx pointer, poly_int64 amount)
 				    next, amount);
 }
 
+/* Return a new RTX holding the result of moving POINTER forward by the
+   size of the mode it points to.  */
+
+static rtx
+arc64_progress_pointer (rtx pointer)
+{
+  return arc64_move_pointer (pointer, GET_MODE_SIZE (GET_MODE (pointer)));
+}
+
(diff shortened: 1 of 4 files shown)


gcc/config/arc64/arc64.c
```diff
@@ -4705,6 +4734,89 @@ arc64_simd128_split_move (rtx *operands, machine_mode mode)
...
+arc64_expand_cpymem (rtx *operands)
+{
+  int n, mode_bits;
+  rtx dst = operands[0];
+  rtx src = operands[1];
+  rtx base;
+  machine_mode cur_mode = BLKmode, next_mode;
+  bool speed_p = !optimize_function_for_size_p (cfun);
+
+  //const int max_num_moves =
```
claziss: C++ style comment, please remove

gcc/config/arc64/arc64.c
```diff
@@ -4705,6 +4734,89 @@ arc64_simd128_split_move (rtx *operands, machine_mode mode)
...
+  //  TARGET_WIDE_LDST ? 2 * MOVE_RATIO (speed_p) : MOVE_RATIO (speed_p);
```
claziss: Likewise

shahab-vahedi:
@claziss ping.

PR #103, opened 2021-12-24
[RFC]  arc: libgcc: fix unwind after signal handler

geomatsi:
ARC specific MD_FALLBACK_FRAME_STATE_FOR handler in libgcc unwind code stores return address in the 31st column of regs array in the frame state structure. However by doing so it rewrites restored blink register value of the function interrupted by a signal. As a result, libgcc may fail to unwind beyond the interrupted function. Usually it does not happen since most functions keep blink value on stack. So libgcc unwinder obtains that value from FDE record later on. However if blink is not on FDE, unwinder fails to locate the next function in the callstack. Moreover, if FDE is empty, unwinder starts to loop over the same function and never completes. Suchb example with uClibc pthreads cancellation is provided further.

This problem affects both ARC700 and ARCv2. However for ARCv2 it is hidden by another issue with __default_rt_sa_restorer implementation in uClibc: see https://github.com/foss-for-synopsys-dwc-arc-processors/uClibc/pull/6

Because of that issue, libgcc unwinder fail to recognize signal frame correclty and completes too early, before it stumbles upon the issue in question.

Suggested fix is to store the return address from signal handler not in blink register position. But keep it in any other register position that is not used in Linux by both ARC700 and ARCv2. This change suggests to use 29 (ilink) for this purpose.

Here is an example where libgcc unwinder ends up in the endless loop. Consider the following simple worker thread:

```
static void* thread_loop(void* args)
{
        for (i = 0; ; i++) {
                printf("worker: loop %d\n", i);
                sleep(1);
        }
}
```

In the case of uClibc, nanosleep syscall is called from the following function:

```
000155a8 <__nanosleep_nocancel>:
   155a8:       208a 1941               mov     r8,101
   155ac:       226f 003f               trap0
   155b0:       208c 8030               cmp     r0,-1024
   155b4:       20e0 07ce               jls     [blink]
   155b8:       1cfc b7c8               st.aw   blink,[sp,-4]
   155bc:       0c2e ffcf               bl      -980    ;151e8 <__syscall_error>
   155c0:       1404 341f               ld.ab   blink,[sp,4]
   155c4:       7ee0                    j_s     [blink]
   155c6:       78e0                    nop_s
```

Its FDE record looks as follows:

```
000001a0 00000010 000001a4 FDE cie=00000000 pc=000155a8..000155c6
  DW_CFA_nop
  DW_CFA_nop
  DW_CFA_nop
```

When main thread attempts to cancel this worker thread, NPTL uClibc code sends SIGCANCEL signal. In its turn signal handler calls libgcc function _Unwind_ForcedUnwind. However libgcc unwinder fails to pass further than __nanosleep_nocancel function. As a result main thread waits forever on pthread_join.

Regards,
Sergey

### DIFF
--- a/libgcc/config/arc/linux-unwind.h
+++ b/libgcc/config/arc/linux-unwind.h
@@ -120,10 +120,10 @@ arc_fallback_frame_state (struct _Unwind_Context *context,
 	= ((_Unwind_Ptr) &(regs[i])) - new_cfa;
     }
 
-  fs->regs.reg[31].how = REG_SAVED_VAL_OFFSET;
-  fs->regs.reg[31].loc.offset = ((_Unwind_Ptr) (regs[ret])) - new_cfa;
+  fs->regs.reg[29].how = REG_SAVED_VAL_OFFSET;
+  fs->regs.reg[29].loc.offset = ((_Unwind_Ptr) (regs[ret])) - new_cfa;
 
-  fs->retaddr_column = 31;
+  fs->retaddr_column = 29;
 
   return _URC_NO_REASON;
 }


geomatsi:
By the way, for now I posted this PR for dev branch. Let me know what should be the correct gcc branch.

cupertinomiranda:
Again, nothing should be changed unless we have before and after executions of glibc and uclibc testsuites, for these particular changes.
I am sorry, but I have spent way too long on fixing unwinding in the past for accepting these changes lightly.
If it is broken today, we should identify what broke it.

geomatsi:
The first results are for uClibc tests on ARC700. The system under test is Linux on nSIM built using the following core components:
* gcc arc-2020.09 release
* uClibc latest revision
* Linux kernel v5.6.3

1. uClibc non-nptl test suites
With and without fixes
   ``` 
   Total skipped: 15
   Total failed: 4
   Total passed: 179
   ```
2. uClibc nptl test suite results:
* without fixes
   ``` 
   Total skipped: 0
   Total failed: 15
   Total passed: 219
   ```
* with both uClibc and libgcc fixes
   ```
   Total skipped: 0
   Total failed: 2
   Total passed: 232
   ```

Taking into account that the fix concerns signal handler stack unwinding, it looks like it also makes sense to run dejagnu g++ and libstdc++ tests on Linux to verify exception handling. Besides, I guess testing should focus on ARCv2/ARCv3 rather than on ARC700. 

Am I missing something ?

Regards,
Sergey

cupertinomiranda:
Since the patch also impacts glibc and all other targets hs/em, we cannot skip it as well.
Can you make a delta between those results, and share ?
I am curious to see which tests are we talking about. In the past I have fixed unwinding for both uclibc and glibc, and so I would expect unwinding related tests to be all passing, maybe not for ARC700.

If cancel tests are the ones in the delta, then I will be interested to understand which commit made it break since it was working in the past.

cupertinomiranda:
Also, since this is a commit to gcc, we need to precisely understand the problem in order to validate if the presented patch is the proper solution.
Maybe we can discuss it in a talk later once Claudiu and Shahab get back from vacations.

geomatsi:
Hi Cupertino,

Here is a summary of tests for ARCHS on nSIM with the same BSP and toolchain versions as for ARC700 above:
1. with uclibc [fix](https://github.com/foss-for-synopsys-dwc-arc-processors/uClibc/pull/6) and without libgcc fix
  ``` 
  Total skipped: 15
   Total failed: 22
   Total passed: 395
   ````
2. with both uclibc [fix](https://github.com/foss-for-synopsys-dwc-arc-processors/uClibc/pull/6) and libgcc fix
  ``` 
  Total skipped: 15
   Total failed: 12
   Total passed: 405
   ````

Both test logs are attached: 
* [archs-uclibc-tests-nofix.log](https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/files/7781466/archs-uclibc-tests-nofix.log)
* [archs-uclibc-tests-fix.log](https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/files/7781465/archs-uclibc-tests-fix.log)

As in the case with ARC700, it looks like significant part of those failures belongs to pthread cancel tests. Note that the first test results without uClibc fix (in other words, w/o both fixes) would be better since uClibc issue masks unwind issue on ARCHS.

Regards,
Sergey

libgcc/config/arc/linux-unwind.h
```diff
@@ -120,10 +120,10 @@ arc_fallback_frame_state (struct _Unwind_Context *context,
 	= ((_Unwind_Ptr) &(regs[i])) - new_cfa;
     }
 
-  fs->regs.reg[31].how = REG_SAVED_VAL_OFFSET;
-  fs->regs.reg[31].loc.offset = ((_Unwind_Ptr) (regs[ret])) - new_cfa;
+  fs->regs.reg[29].how = REG_SAVED_VAL_OFFSET;
+  fs->regs.reg[29].loc.offset = ((_Unwind_Ptr) (regs[ret])) - new_cfa;
 
-  fs->retaddr_column = 31;
+  fs->retaddr_column = 29;
```
vineetgarc, line 126: In ARC700, ilink is a privilege mode register, you can't write to it.
In ARCHS it is also not available as a GPR.

my bad, you are right, you are just using this as a placeholder and not really writing to the reg

vineetgarc: So this fix might be needed for some cases.
The routine above is called nanosleep_nocancel - does this mean it is supposed to NOT ne cancelled.
At any rate see if you need to add cfi pseudo ops to this routine, that should be the first fix

vineetgarc: Also this is all good analysis and such, but what exactly is the motivation of this change, tracking uClibc failures ? That was something closed a long time ago, whats the sudden interest.

geomatsi: In uClibc nanosleep call is nanosleep_nocancel wrapped by LIBC_CANCEL_ASYNC/LIBC_CANCEL_RESET pair. That is why you see nanosleep_nocancel in disassembler.

geomatsi: IIUC adding `.cfi` might be necessary: I don't understand why they are not added for `st.aw` and `ld.ab` calls. 
However adding `.cfi` is sill not enough for proper unwind since blink value is pushed to the stack after system call. So array contents at blink position should not be rewritten.

geomatsi: > Also this is all good analysis and such, but what exactly is the motivation of this change, tracking uClibc failures ? That was something closed a long time ago, whats the sudden interest.

The reason to investigate that was customer issue on arc700 and pthread cancellation. Finding that archs is also impacted was a byproduct of that investigation.

vineetgarc:
Does this problem affect only uClibc or glibc too ?

geomatsi:
No, according to my tests, glibc is not affected. IIUC the reason is as follows.

In glibc unwind procedure after the signal handler stack frame processing is not correct anyway. But unlike uclibc, it does not fall into the endless loop and finally completes. The reason why it completes is that FDE entries are usually not empty. So if FDE is not empty, unwind process loops over the same interrupted function, but adds non-zero frame size on each iteration. So finally unwind is stopped by the second condition in unwind_stop callback in glibc (see nptl/unwind.c):
```
  if ((actions & _UA_END_OF_STACK)
      || ! _JMPBUF_CFA_UNWINDS_ADJ (buf->cancel_jmp_buf[0].jmp_buf, context,
                                    adj))
    do_longjump = 1;

```

Note that similar behavior is observed in uClibc for other syscalls, that have non-empty FDEs. For instance, if sleep is implemented as select(0, NULL, NULL, NULL, ts). In this case the similar second condition triggers from unwind_stop callback in libpthread/nptl/unwind.c.

claziss:
Hi, the issue found in libgcc's unwinding routine is true. However, the fix is not really what is needed. GCC is using an alternative DWARF2 return column for unwinding a signal frame. 
If you can provide here a simple test that exposes this issue, I can try to craft a proper patch.
Thanks.

geomatsi:
> Hi, the issue found in libgcc's unwinding routine is true. However, the fix is not really what is needed. GCC is using an alternative DWARF2 return column for unwinding a signal frame. If you can provide here a simple test that exposes this issue, I can try to craft a proper patch. Thanks.

You may use the attached simple [pthread cancel](https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/files/7809700/test.txt) example mentioned in the PR description. Use any release uClibc toolchain for arc700 or custom uClibc toolchain for archs with [patched uClibc](https://github.com/foss-for-synopsys-dwc-arc-processors/uClibc/pull/6). 

Alternatively, you may just attach or post the patch here. Then I will help with testing.

claziss:
Pushed a patch on the topic https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/commit/6427669a55d1fe364be7cffbaf67cb0a707a1cfb
If alright, please close this PR.

geomatsi:
> Pushed a patch on the topic [6427669](https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/commit/6427669a55d1fe364be7cffbaf67cb0a707a1cfb) If alright, please close this PR.

I completed uclibc tests for builds with uClibc fix and new proper libgcc fix on both arc700 (nSIM) and archs (HSDK). Everything is alright: I see the same improvements of test results, in particular pthread cancel tests. 

Thanks for the proper libgcc fix! Closing down this PR and connected issues.

Regards,
Sergey

geomatsi:
For the sake of completeness, here is a delta between dejagnu tests on HSDK with and without uClibc and gcc fixes: 
[dejagnu-delta.json.txt](https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/files/7856481/dejagnu-delta.json.txt)

Summary:
* no changes for glibc toolchain
* changes for uClibc toolchain:
  * g++: fixed cleanup-8, cleanup-9, cleanup-10, cleanup-11 tests
  * gcc: fixed cleanup-8, cleanup-9, cleanup-10, cleanup-11 tests

PR #104, opened 2022-01-19
arc64: Change modifier %h to %H, and %H to %V

geomatsi:
Hi @shahab-vahedi , @cupertinomiranda 

I have been working on Buldroot for arc32. Buildroot for arc64 uses arc64 branches for gcc/binutils/glibc. Following arc-gnu-toolchain scripts I used arc32 branches for arc32. That works fine, the only missing thing for now is this gcc commit by Claudiu.
Having this commit in arc32 branch simplifies the work on Linux bring-up for HS5x CPU: in certain cases we can use the same generic inline assembly across HS4x/HS5x/HS6x targets.

Regards,
Sergey


shahab-vahedi (approved):
All's good! sorry for the delay.

PR #105, opened 2022-01-26
ARC32: fixes for large mcmodel

geomatsi:
Cherry-pick fixes for large mcmodel from arc64 branch to arc32 branch. With these changes we now have working Linux kernel modules on HS5x.


shahab-vahedi:
I've rebased the `arc32` branch onto the `arc64` branch. That should have picked up all the cherry-picked commits. Therefore no need for this PR anymore. Thank you for bringing it into our attention.

PR #121, opened 2024-06-20
ARCv2: Update pr9001184797.c test case.

luismgsilva:
Updated the test case to comply with new standards due to stricter analysis in the latest GCC versions.


shahab-vahedi (approved):

shahab-vahedi:
@luismgsilva Could you upstream this as well?

PR #122, opened 2024-06-21
libstdc++: Check for fenv.h and complex.h earlier

kolerov:
Initially this check was performed before `GLIBCXX_CHECK_C99_TR1` to prevent detecting `fenv.h` and `complex.h` from C++ headers. But since GCC 14 `GLIBCXX_ENABLE_C99` also tries to check `fenv.h` and finds it C++ headers even if it is not presented in `libc`. Thus, the check must be performed earlier.

### DIFF
--- a/libstdc++-v3/configure
+++ b/libstdc++-v3/configure
@@ -16585,6 +16585,24 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
 
 
 
+# Check for fenv.h and complex.h before GLIBCXX_CHECK_C99_TR1
+# and GLIBCXX_ENABLE_C99 so that the check is done with the C compiler
+# (not C++). Checking with C++ can break a canadian cross build if either
+# file does not exist in C but does in C++.
+for ac_header in fenv.h complex.h
+do :
+  as_ac_Header=`$as_echo "ac_cv_header_$ac_header" | $as_tr_sh`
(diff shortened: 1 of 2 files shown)


artemiy-volkov:
patch itself is good, but some improvements to the commit message are in order, e.g.
s/Initially/Initially,
s/thise/this
s/presented/present
s/since/, since
s/. Thus//

additionally, I would start the message by mentioning that you're changing configure scripts, I think that makes it easier to put the rest into context

kolerov:
@artemiy-volkov I've updated the commit.

artemiy-volkov:
> @artemiy-volkov I've updated the commit.

Please merge these two sentences:

But, since GCC 14 GLIBCXX_ENABLE_C99 also tries to check fenv.h
and finds it C++ headers even if it is not present in libc.
The check must be performed earlier.

and add "in" before "C++ headers" to fix the grammar:

But, since GCC 14 GLIBCXX_ENABLE_C99 also tries to check fenv.h
and finds it in C++ headers even if it is not present in libc, the check must be performed earlier.

LGTM with that change!

kolerov:
@artemiy-volkov Done!

artemiy-volkov (approved):

shahab-vahedi (approved):
This is worth upstreaming Yuriy.

Also, this must be pushed onto `arc64` branch as well.

shahab-vahedi:
@kolerov where is the upstream patch submission?

PR #125, opened 2024-10-01
RHX-100 scheduling (initial)

artemiy-volkov:
RHX-100 scheduling as well as assorted middle-end fixes.


dobbelaj-snps:
I assume we will need tests added as well ? It is ok if we do that in a separate merge request.

Alexehv77 (approved):
Looks fine to me. 
Would be good to make tickets where fused instructions arent matched.

PR #128, opened 2024-10-22
arc64: do not emit custom C++ thunks under large code model

artemiy-volkov:
Currently, under -mcmodel=large, arc64_output_mi_thunk () always tries to create a new pseudo to accommodate a long call, which leads to a post-reload crash even for a very simple diamond inheritance pattern.  To fix this, define arc64_can_output_mi_thunk () returning false in this scenario, resulting in default machinery being used as opposed to target-specific thunks. Additionally, create a new gcc.target testcase checking that this patch fixes the issue.


artemiy-volkov:
preempted by #129

PR #129, opened 2024-10-24
arc64: fix for multiple inheritance of virtual functions

Alexehv77:
Fix regarding thunk code generated for C++ virtual calls with multiple inheritance. The thunk is a wrapper code that is emitted after reload, thus no free registers are available.


luismgsilva (commented):
All tests passed.

luismgsilva (approved):

gcc/config/arc64/arc64.cc
```diff
@@ -2826,6 +2828,13 @@ arc64_output_mi_thunk (FILE *file,
       emit_insn (gen_add3_insn (this_rtx, this_rtx, temp1));
     }
 
+  if (arc64_is_long_call_p (temp0) && !REG_P (temp0)){
```
artemiy-volkov: Please check your patch for code style using the "./contrib/check_GNU_style.sh $patchfile" script (see also the first 2 lines of https://gcc.gnu.org/wiki/FormattingCodeForGCC)

gcc/config/arc64/arc64.cc
```diff
@@ -2777,17 +2777,19 @@ arc64_output_mi_thunk (FILE *file,
 		       tree function)
 {
   const char *fnname = IDENTIFIER_POINTER (DECL_ASSEMBLER_NAME (thunk_fndecl));
-  rtx this_rtx, fnaddr, temp1;
+  rtx this_rtx, fnaddr, temp0, temp1;
   rtx_insn *insn;
-
+  machine_mode mode;
```
artemiy-volkov, line 2783: why was the empty line deleted?

gcc/config/arc64/arc64.cc
```diff
@@ -5322,7 +5331,15 @@ arc64_expand_call (rtx result, rtx mem, bool sibcall)
      address of the callee into a register before performing the
      branch-and-link.  */
   if (arc64_is_long_call_p (callee) && !REG_P (callee))
-    XEXP (mem, 0) = force_reg (mode, callee);
+    if (can_create_pseudo_p())
```
artemiy-volkov: please surround this if-statement with braces, since it's quite long

gcc/config/arc64/arc64.md
```diff
@@ -1189,11 +1189,12 @@ xorl"
 
 ;FIXME! add short variant for jump
 (define_insn "*sibcall<mode>_insn"
-  [(call
+ [(parallel [(call
```
artemiy-volkov: Could you please explain this change? Is it really necessary? See p.2, paragraph 2 here https://gcc.gnu.org/onlinedocs/gccint/Patterns.html:
If the vector has multiple elements, the RTL template is treated as a parallel expression.

gcc/config/arc64/arc64.h
```diff
@@ -302,7 +302,7 @@ enum reg_class
 {								\
   { 0x00000000, 0x00000000, 0x00000000, 0x00000000 }, /* NO_REGS */	\
   { 0x0000f00f, 0x00000000, 0x00000000, 0x00000000 }, /* AC16_REGS */	\
-  { 0x00001fff, 0x00000000, 0x00000000, 0x00000000 }, /* SIBCALL_REGS */ \
+  { 0x00003fff, 0x00000000, 0x00000000, 0x00000000 }, /* SIBCALL_REGS */ \
```
artemiy-volkov, line 305: This change isn't reflected in the commit message and isn't very obvious without an explanation.

gcc/config/arc64/arc64.cc
```diff
@@ -5322,7 +5331,15 @@ arc64_expand_call (rtx result, rtx mem, bool sibcall)
...
+      XEXP (mem, 0) = force_reg (mode, callee);
+    else
+      /* Normally at this point no registers can be used for allocation.
+       * There are special casses where one can safely use registers e.g.,
```
artemiy-volkov: s/casses/cases

artemiy-volkov (changes requested):
Thanks for the patch, could you please:

(1) Add 1-3 sentences to the commit message clarifying the implementation details (used R13 for sibcall since it's caller-saved, used it in arc64_output_mi_thunk () etc. etc.), also mentioning the testcase. This sets up the context for review nicely.
(2) Fix formatting errors reported by ./contrib/check_GNU_style.sh
(3) Fix the remaining comments

Many thanks in advance

gcc/config/arc64/arc64.md
Alexehv77: Done

gcc/config/arc64/arc64.h
Alexehv77: I now refer in the commit message about the R13

gcc/config/arc64/arc64.cc
Alexehv77: Done

gcc/config/arc64/arc64.cc
Alexehv77: Done

gcc/config/arc64/arc64.cc
Alexehv77: Done

gcc/config/arc64/arc64.cc
Alexehv77: Done

gcc/config/arc64/arc64.md
```diff
@@ -1189,7 +1189,7 @@ xorl"
 
 ;FIXME! add short variant for jump
 (define_insn "*sibcall<mode>_insn"
-  [(call
+ [(call
```
artemiy-volkov: this has no functional value, so please just drop this chunk

gcc/testsuite/gcc.target/arc64/cmodel-large-cpp-thunk.cpp
```diff
@@ -0,0 +1,26 @@
...
+  virtual void bar() = 0;
+};
+class C : public A, public B {
+  NO_INLINE void foo() {}
+  NO_INLINE void bar() {}
+};
+
+C c;
+
+int main() {
```
artemiy-volkov, line 22: Could we make this a run test in addition to compile? You'd need to make foo() and bar() output something, call those functions from an instance of class C, then use the "dg-do run" and "dg-output" directives to check for correctness. For an example of the latter, see gcc/testsuite/c-c++-common/ubsan/div-by-zero-1.c.

artemiy-volkov (changes requested):
Could you please fix the remaining ./contrib/check_GNU_style.sh complaints? (there's 1 false positive I think)

=================================================
artemiy@us01arcgnu3:/SCRATCH/art/src/gcc-classic $ git checkout turjan/long-call-in-thunk-fix
branch 'turjan/long-call-in-thunk-fix' set up to track 'origin/turjan/long-call-in-thunk-fix'.
Switched to a new branch 'turjan/long-call-in-thunk-fix'
artemiy@us01arcgnu3:/SCRATCH/art/src/gcc-classic $ git format-patch -1
0001-arc64-fix-for-multiple-inheritance-of-virtual-functi.patch
artemiy@us01arcgnu3:/SCRATCH/art/src/gcc-classic $ ./contrib/check_GNU_style.sh 0001-arc64-fix-for-multiple-inheritance-of-virtual-functi.patch

Lines should not exceed 80 characters.
63:+      /* Make use of R13_REGNUM to store the sibcall address. This register is caller saved

Dot, space, space, new sentence.
63:+      /* Make use of R13_REGNUM to store the sibcall address. This register is caller saved

Dot, space, space, end of comment.
64:+     and can be used locally as a temporary register.*/

Sentences should end with a dot.  Dot, space, space, end of the comment.
101:+  { 0x00003fff, 0x00000000, 0x00000000, 0x00000000 }, /* SIBCALL_REGS */ \\

There should be exactly one space between function name and parenthesis.
48:+  mode = GET_MODE(temp0);

===================================================

Good to go with those changes and the 2 pending comments in the review. Thanks!

gcc/testsuite/gcc.target/arc64/cmodel-large-cpp-thunk.cpp
Alexehv77: There were added 2 test files - one that checks the assembly and the other the execution.

PR #131, opened 2024-11-04
arcv: add next_insn existence checks to some scheduler hooks

artemiy-volkov:
In arcv_macro_fusion_pair_p () and riscv_sched_reorder2 (), there is null pointer dereference potential when doing things with next_insn (insn) without first checking that it exists.  Rectify that by adding such checks where necessary.


gcc/config/riscv/riscv.cc
Alexehv77, line 1: To me your changes are fine.

PR #133, opened 2024-12-05
arcv: adjust scheduling priority for load/store pairs

artemiy-volkov:
This patch implements riscv_sched_adjust_priority () for the RHX-100 microarchitecture by slightly bumping the priority of load/store pairs.  As a consequence of this change, it becomes easier for riscv_sched_reorder2 () to schedule instructions in the memory pipe.

### DIFF
--- a/gcc/config/riscv/riscv.cc
+++ b/gcc/config/riscv/riscv.cc
@@ -9427,6 +9427,26 @@ riscv_sched_adjust_cost (rtx_insn *insn, int dep_type, rtx_insn *dep_insn, int c
   return new_cost;
 }
 
+static int
+riscv_sched_adjust_priority (rtx_insn *insn, int priority)
+{
+  if (!riscv_is_micro_arch (arcv_rhx100))
+    return priority;
+
+  /* Bump the priority of fused load-store pairs for easier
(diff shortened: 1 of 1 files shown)


gcc/config/riscv/riscv.cc
```diff
@@ -9427,6 +9427,23 @@ riscv_sched_adjust_cost (rtx_insn *insn, int dep_type, rtx_insn *dep_insn, int c
   return new_cost;
 }
 
+static int
+riscv_sched_adjust_priority (rtx_insn *insn, int priority)
+{
+  if (!riscv_is_micro_arch (arcv_rhx100))
+    return priority;
+
```
Alexehv77, line 9435: you may write an explanation why the first instruction within 2 such couples needs higher priority. perhaps has to do with the chance to fuse them.

artemiy-volkov: Added a comment.

Alexehv77 (approved):
For future keep in mind that finding a way to let the compiler to calculate the priority based on path criticality, resources utilization or number of consumers, may improve the performance. Current implementation does it blindly on all the ld/ld or st/st which most likely has drawbacks.

PR #139, opened 2025-02-10
arcv: Remove mtune=arc-v-rmx-100-series from t-arcv config list

### LINKED ISSUE #656: GCC does not link binaries with default targets options
This issue is presented in `arc-2024.12` release and may be reproduced this way:

```
$ riscv64-snps-elf-gcc test.c
riscv64-snps-elf-gcc: fatal error: Cannot find suitable multilib set for '-march=rv64imafdc_zicsr_zifencei'/'-mabi=lp64d'
compilation terminated.
```

If `-mtune=rocket` is passed, then a binary is linked successfully:

```
$ riscv64-snps-elf-gcc test.c -mtune=rocket
```

The problem is in how GCC for RISC-V works with multilib configurations. Let's start from the ground.

GCC for ARC-V has its own custom multilib [configurations file](https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/blob/arc-2024.12/gcc/config/riscv/t-arcv), which contains a list of all prebuilt configurations. Let's take a few lines:

```
MULTILIB_REQUIRED   += march=rv32i/mabi=ilp32
MULTILIB_REQUIRED   += march=rv64imafdc/mabi=lp64d/mtune=arc-v-rmx-100-series/mcmodel=medany
```

Internally each configuration is described by [`riscv_multi_lib_info_t` structure](https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/blob/arc-2024.12/gcc/common/config/riscv/riscv-common.cc#L437):

```c
struct riscv_multi_lib_info_t {
  std::string path;
  std::string arch_str;
  std::string abi_str;
  std::vector<std::string> conds;
  riscv_subset_list *subset_list;

  static bool parse (struct riscv_multi_lib_info_t *,
		     const std::string &,
		     const std::vector<std::string> &);
};
```

For `march=rv64imafdc/mabi=lp64d/mtune=arc-v-rmx-100-series/mcmodel=medany` configuration it may look like this:

```python
path = "rv64imafdc/lp64d/rmx100/medany",
arch_str = "rv64imafdc",
abi_str = "lp64d",
conds = {
    "march=rv64imafdc",
    "mabi=lp64d",
    "mtune=arc-v-rmx-100-series",
    "!mtune=arc-v-rmx-500-series",
    "!mtune=arc-v-rhx-100-series",
    "mcmodel=medany",
},
subset_list = "rv64i2p1_m2p0_a2p1_f2p2_d2p2_c2p0_zicsr2p0"
```

When GCC decides what configuration to choose, it goes through all configurations and compares them with a combination of passed options. GCC compares `mabi` (`abi_str`) first, then it compares `march` (`subset_list`). Then it compares other conditionals (`conds`), which were enlisted in `MULTILIB_OPTIONS` in `t-arcv` and were presented in configuration's string: `march` (yet again), `mabi` (yet again too), `mtune` and `mcmodel`. So, this sample configuration is selected only when `march`/`mabi` are compatible and only when `mtune=arc-v-rmx-100-series` and `mcmodel=medany` are passed.

Everything becomes a little bit weird when you want to build an application with a default list of options without passing anything to GCC. In case of `arc-2024.12` it corresponds to `-march=rv64imafdc_zicsr_zifencei -mabi=lp64d -mtune=arc-v-rmx-100-series`. For this default configuration `riscv_multi_lib_info_t` structure looks like this:

```python
path = ".",
arch_str = "rv64imafdc_zicsr_zifencei",
abi_str = "lp64d",
conds = {
    "!mtune=arc-v-rmx-100-series",
    "!mtune=arc-v-rmx-500-series",
    "!mtune=arc-v-rhx-100-series",
    "!mcmodel=medany",
},
subset_list = "rv64i2p1_m2p0_a2p1_f2p2_d2p2_c2p0_zicsr2p0_zifencei2p0"
```

Despite the fact that a default configuration is built with `-mtune=arc-v-rmx-100-series` and `arc-v-rmx-100-series` is a default value for `-mtune`, you cannot choose a default configuration by passing default set of options. It happens because formally that default configuration does not have `mtune` selector in a final multilib configuration file (generated from `t-arcv`) and GCC assumes that all `mtune` variants enlisted in `t-arcv` are not applicable for this default configuration.

However, when `-mtune=rocket` is passed (see an example in the beginning of the issue), all `mtune` conditions for the default configuration are passed successfully.

kolerov:
If GCC is built with --with-mtune=arc-v-rmx-100-series and mtune=arc-v-rmx-100-series is mentioned in MULTILIB_OPTIONS list of multilib options, then GCC cannot link an application properly in these cases:

1. When GCC is used without any multilib options with default options only.
2. When a multilib configuration does not involve mtune, e.g. march=rv64imafdc/mabi=lp64d/mcmodel=medany.

In both cases -mtune=arc-v-rmx-100-series is passed explicitly, but the corresponding configuration cannot be chosen since mtune=arc-v-rmx-100-series is not mentioned in t-arcv for them.

Since arc-v-rmx-100-series is a default -mtune value for ARC-V, then mtune=arc-v-rmx-100-series maybe omitted in t-arc and assumed to be a default -mtune value for all configurations without mtune=... field.

This PR fixes https://github.com/foss-for-synopsys-dwc-arc-processors/toolchain/issues/656.

### DIFF
--- a/gcc/config/riscv/t-arcv
+++ b/gcc/config/riscv/t-arcv
@@ -58,24 +58,21 @@ MULTILIB_DIRNAMES   += $(arcv_march_variants)
 MULTILIB_OPTIONS    += mabi=ilp32e/mabi=ilp32/mabi=ilp32f/mabi=ilp32d/mabi=lp64/mabi=lp64d
 MULTILIB_DIRNAMES   +=      ilp32e      ilp32      ilp32f      ilp32d      lp64      lp64d
 
-MULTILIB_OPTIONS    += mtune=arc-v-rmx-100-series/mtune=arc-v-rmx-500-series/mtune=arc-v-rhx-100-series
-MULTILIB_DIRNAMES   +=       rmx100                     rmx500                     rhx100
+MULTILIB_OPTIONS    += mtune=arc-v-rmx-500-series/mtune=arc-v-rhx-100-series
+MULTILIB_DIRNAMES   +=       rmx500                     rhx100
 
 MULTILIB_OPTIONS    += mcmodel=medany
 MULTILIB_DIRNAMES   +=         medany
 
-# Fallback configurations for general targets
-MULTILIB_REQUIRED   += march=rv32i/mabi=ilp32
-
 # Fallback configurations for RMX-100 mini
-MULTILIB_REQUIRED   += march=rv32e/mabi=ilp32e/mtune=arc-v-rmx-100-series
+MULTILIB_REQUIRED   += march=rv32e/mabi=ilp32e
 
 # Fallback configurations for RMX-100
-MULTILIB_REQUIRED   += march=rv32i/mabi=ilp32/mtune=arc-v-rmx-100-series
-MULTILIB_REQUIRED   += march=rv32ia/mabi=ilp32/mtune=arc-v-rmx-100-series
-MULTILIB_REQUIRED   += march=rv32imc_zicsr/mabi=ilp32/mtune=arc-v-rmx-100-series
-MULTILIB_REQUIRED   += march=rv32iac_zicsr/mabi=ilp32/mtune=arc-v-rmx-100-series
-MULTILIB_REQUIRED   += march=rv32imac_zicsr/mabi=ilp32/mtune=arc-v-rmx-100-series
+MULTILIB_REQUIRED   += march=rv32i/mabi=ilp32
+MULTILIB_REQUIRED   += march=rv32ia/mabi=ilp32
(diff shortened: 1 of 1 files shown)


artemiy-volkov (commented):
I don't like this since we are now dependent on the user configuring with --with-tune=rmx100. I don't think rmx100 should get any special treatment here, this is an extremely fragile approach

kolerov:
An alternative is to use `rocket` as the default `mtune` value. But in this case `--mtune=arc-v-*-series` must be always passed explicitly to GCC to link with ARC-V specific configurations (with `mtune` value in `t-arcv`).

PR #141, opened 2025-03-10
arcv: handle PARALLELs and CLOBBERs in arcv_propagate_hard_register_c…

artemiy-volkov:
…opies ()

Currently, the ARC-V-specific machine-reorg hardreg propagation pass handles only single_set's, leading to subtle errors when hardregs are either clobbered or updated inside a PARALLEL rtx. Fix this by adding handling logic for CLOBBER and PARALLEL expressions. The behavior is simple: clear all information pertaining to the clobbered or parallel-set regs from the reg_originals array.

### DIFF
--- a/gcc/config/riscv/riscv.cc
+++ b/gcc/config/riscv/riscv.cc
@@ -10585,57 +10585,98 @@ arcv_propagate_hard_register_copies (void)
 		reg_originals[i] = 0;
 	    }
 
-	  set = single_set (insn);
-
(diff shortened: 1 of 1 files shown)


Alexehv77 (approved):
I understand the fix as taking care of copy propagation in the target dependent last step - reorg. Would be also possible for later to investigate the opportunity of running hard reg cprop. This could be done such that is triggered only for riscv so it wont affect other targets.

PR #145, opened 2025-03-24
arcv: introduce --param=arcv-ld-cycles

artemiy-volkov:
Since it is critical to scheduling for the RMX-100 core, allow setting load latency via a newly create target-specific arcv-ld-cycles parameter. The worst case latency is 3, and so is the parameter's default value.


Alexehv77:
for me is fine

PR #146, opened 2025-03-25
arcv: add a multilib config without Zc* for RMX-100

artemiy-volkov:
Since compressed instructions are no longer efficient on the RMX-100, add a rv32im_zb*_zicsr fallback configuration for that core.


abrodkin:
> Since compressed instructions are no longer efficient on the RMX-100

What does that really mean?

artemiy-volkov:
> > Since compressed instructions are no longer efficient on the RMX-100
> 
> What does that really mean?

Due to recent uarch changes, a jump to an instruction unaligned to 32 bits incurs a performance penalty. Removing all compressed instructions is one way to deal with that limitation, at the cost of increased code size.

PR #153, opened 2025-05-16
arcv: implement the XARCVudsp extension

artemiy-volkov:
This commit implements support for the Synopsys ARC-V micro-DSP extension, version 1.4.  The extension is only available for 32-bit targets and consists of a small number of DSP operations on 16-bit and 32-bit data.  Intrinsics added as part of this patch are designed to resemble RVV intrinsics (with severe limitations to SEW and LMUL combinations allowed for each instruction), but are actually implemented similarly to other DSP-style extensions such as Xvcalu.

FIXME: add more pointers to public documentation.


Alexehv77 (commented):
As a general comment as you added few dozen of new instructions supported by intrinsics would be useful to add dejagnu tests that at least validate the compilation process. I guess you already have such C code that you used during the implementation - so you can just added to the already existent tests. If you plan to do it in a subsequent patch then is no problem to approve it.

artemiy-volkov:
> As a general comment as you added few dozen of new instructions supported by intrinsics would be useful to add dejagnu tests that at least validate the compilation process. I guess you already have such C code that you used during the implementation - so you can just added to the already existent tests. If you plan to do it in a subsequent patch then is no problem to approve it.

Thanks for your review so far, tests have been added and validated.

Alexehv77 (approved):

PR #155, opened 2025-05-28
arc: Use intrinsics for __builtin_mul_overflow () 

luismgsilva:


luismgsilva:
@artemiy-volkov At the time, I had to revert a patch from Roger Sayle. Now, how should we proceed? Revert the revert?

artemiy-volkov:
> @artemiy-volkov At the time, I had to revert a patch from Roger Sayle. Now, how should we proceed? Revert the revert?

Yes, that would be the best for now. Both the revert and the reapplication will disappear in the 2025.12 branch.

PR #156, opened 2025-06-04
ARC-V DSP extensions

artemiy-volkov:


luismgsilva:
Looks good to me so far. There are some minor issues with the GNU style, but I assume it will either be fixed, or not important.

gcc/config/riscv/arcv-vector.md
```diff
@@ -0,0 +1,2438 @@
...
+	  UNSPEC_ARCV_VCLR)
+	(match_operand:V_VLSI 2 "vector_merge_operand"     "vu,0,vu,0,vu,0,vu,0,vu,0,vu,0")))]
+  "TARGET_XARCVVDSP"
+  "arcv.vclr.v\t%0,%3,%4%p1"
+  [(set_attr "type" "viwmuladd")
+   (set_attr "mode" "<MODE>")
+  (set_attr "vl_op_idx" "4")
+  (set (attr "ta") (symbol_ref "riscv_vector::get_ta(operands[5])"))
+  (set (attr "ma") (symbol_ref "riscv_vector::get_ma(operands[6])"))
+  (set (attr "avl_type_idx") (const_int 7))])
```
MichielDerhaeg: Indentation is not consistent.

MichielDerhaeg (commented):
What should we do with the perl script? Should we put this somewhere in case we need to make fixes/changes? Or do you think we won't need it anymore after we're done? I also don't think it is idempotent, you would have to revert the autogenerated changes to reapply them I think. Would those changes then be separated in a different commit?

luismgsilva (approved):

gcc/config/riscv/arcv-vector.md
artemiy-volkov: Fixed, thank you.

artemiy-volkov:
> What should we do with the perl script? Should we put this somewhere in case we need to make fixes/changes? Or do you think we won't need it anymore after we're done? I also don't think it is idempotent, you would have to revert the autogenerated changes to reapply them I think. Would those changes then be separated in a different commit?

It's up to you what to do with it, the binutils part is fine but the GCC part is not easily maintainable, so my recommendation would be to throw it away.

MichielDerhaeg (approved):
LGTM

PR #158, opened 2025-06-24
arc64: select correct encoding for 'mpyuw b, b, s12'

MichielDerhaeg:
Fix for https://github.com/foss-for-synopsys-dwc-arc-processors/toolchain/issues/663

### DIFF
--- a/gcc/config/arc64/arc64.h
+++ b/gcc/config/arc64/arc64.h
@@ -477,6 +477,7 @@ extern const enum reg_class arc64_regno_to_regclass[];
 #define UNSIGNED_INT8(X) (UNSIGNED(X,8))
 #define UNSIGNED_INT9(X) (UNSIGNED(X,9))
 #define UNSIGNED_INT10(X) (UNSIGNED(X,10))
+#define UNSIGNED_INT11(X) (UNSIGNED(X,11))
 #define UNSIGNED_INT12(X) (UNSIGNED(X,12))
 #define UNSIGNED_INT16(X) (UNSIGNED(X,16))
 // TODO: Fix for 32 bit compiler host architecture.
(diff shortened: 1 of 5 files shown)


MichielDerhaeg:
I wasn't sure whether the instruction actually interpreted the s12 operand as a signed value, so I devised an experiment:
```lang=c
#include <stdio.h>

#define MPYUW(rd, rs, imm) \
  asm("mpyuw %0, %1, %2" \
    : "=r"(rd) \
    : "r"(rs), "i"(imm) \
    : "cc")

int main() {
  unsigned rd;
  MPYUW(rd, 1, 0x7FF);
  printf("0x%x\n", rd);
  MPYUW(rd, 1, -0x800);
  printf("0x%x\n", rd);
  MPYUW(rd, 1, -0x1);
  printf("0x%x\n", rd);
}
```
It produces the desired instructions in the assembled binary:
```lang=asm
     300:	209f 0fdf           	mpyuw	r0,r0,2047
     31a:	209f 0020           	mpyuw	r0,r0,-2048
     344:	209f 0fff           	mpyuw	r0,r0,-1
```
It outputs these values when the binary is executed with nsim.
```
0x7ff
0xf800
0xffff
```
This means the instruction performs a sign extension to 16-bits before doing a unsigned muliplication.

I solved the issue by splitting out the sign extension case as a separate `define_insn` in GCC.

MichielDerhaeg:
I also checked all other instructions that potentially have s12 immediate operands, but none of them are technically incorrect. They usually just don't support the case when the operand is signed (negative), but won't produce incorrect instructions.

MichielDerhaeg:
I still have to figure out testing.

luismgsilva:
LGTM. Please add a description in the commit message reporting the changes and why. Thanks!

luismgsilva (approved):

PR #159, opened 2025-07-03
fixup! arcv: implement the XARCVudsp extension

artemiy-volkov:
Designate the output operand of the LMUL=1/2 variant of arcv.xvwmul.vv as early-clobber to avoid overlap with one of the inputs.


luismgsilva (approved):
LGTM. Thanks

PR #160, opened 2025-07-14
arcv: Add APEX support.

luismgsilva:

### DIFF
--- a/gcc/config/riscv/arcv-apex.md
+++ b/gcc/config/riscv/arcv-apex.md
@@ -44,9 +44,9 @@
 )
 
 ;; Used by "XI","XD" insn. format: `insn src0` volatile
-(define_insn "riscv_arcv_apex_void_ftype_src0_v"
-  [(unspec_volatile:SI [(match_operand:SI 0 "const_int_operand" "xAVpXI,xAVpXD")
-			(match_operand:SI 1 "nonmemory_operand" "I,r")]
+(define_insn "riscv_arcv_apex_void_ftype_<S0M:mode>_v"
+  [(unspec_volatile [(match_operand:SI 0 "const_int_operand" "xAVpXI,xAVpXD")
+			(match_operand:S0M 1 "nonmemory_operand" "I,r")]
 	UNSPEC_ARCV_APEX_VOID_SRC0_V)]
   ""
 {
@@ -68,11 +68,30 @@
   [(set_attr "type" "arith,arith")]
 )
 
+(define_expand "riscv_arcv_apex_void_ftype_src0_v"
+  [(unspec_volatile [(match_operand:SI 0 "const_int_operand")
+    (match_operand 1 "nonmemory_operand")]
+	UNSPEC_ARCV_APEX_VOID_SRC0_V)]
+  ""
+{
+  /* Build the SET exactly as it appears above, but with the
+     real RTX objects.  Every operand already carries its mode,
+     so nothing needs to be guessed.  The UNSPEC code is
+     essential as it tags this RTL with a unique ID, so the
+     recognizer can match it to the correct mode-specific
+     define_insn.  */
+  emit_insn (gen_rtx_UNSPEC_VOLATILE (VOIDmode,
+				 gen_rtvec (2, operands[0],
+					       operands[1]),
+				 UNSPEC_ARCV_APEX_VOID_SRC0_V));
+  DONE;
+})
+
 ;; Used by "XS","XD" insn. format: `insn src0, src1` volatile
-(define_insn "riscv_arcv_apex_void_ftype_src0_src1_v"
-  [(unspec_volatile:SI [(match_operand:SI 0 "const_int_operand" "xAVpXS,xAVpXD")
-			(match_operand:SI 1 "register_operand" "r,r")
-			(match_operand:SI 2 "nonmemory_operand" "B8,r")]
+(define_insn "riscv_arcv_apex_void_ftype_<S0M:mode>_<S1M:mode>_v"
+  [(unspec_volatile [(match_operand:SI 0 "const_int_operand" "xAVpXS,xAVpXD")
+			(match_operand:S0M 1 "register_operand" "r,r")
+			(match_operand:S1M 2 "nonmemory_operand" "B8,r")]
 	UNSPEC_ARCV_APEX_VOID_SRC0_SRC1_V)]
   ""
 {
@@ -96,10 +115,31 @@
   [(set_attr "type" "arith,arith")]
 )
 
+(define_expand "riscv_arcv_apex_void_ftype_src0_src1_v"
+  [(unspec_volatile [(match_operand:SI 0 "const_int_operand")
+    (match_operand 1 "register_operand")
+    (match_operand 2 "nonmemory_operand")]
+	UNSPEC_ARCV_APEX_VOID_SRC0_SRC1_V)]
+  ""
+{
+  /* Build the SET exactly as it appears above, but with the
+     real RTX objects.  Every operand already carries its mode,
+     so nothing needs to be guessed.  The UNSPEC code is
+     essential as it tags this RTL with a unique ID, so the
+     recognizer can match it to the correct mode-specific
+     define_insn.  */
+  emit_insn (gen_rtx_UNSPEC_VOLATILE (VOIDmode,
+				 gen_rtvec (3, operands[0],
+					       operands[1],
+					       operands[2]),
+				 UNSPEC_ARCV_APEX_VOID_SRC0_SRC1_V));
+  DONE;
+})
+
 ;; Used by "XD" insn. format: `insn dest` volatile
-(define_insn "riscv_arcv_apex_dest_ftype_v"
-  [(set (match_operand:SI 0 "register_operand" "=r")
-	(unspec_volatile:SI [(match_operand:SI 1 "const_int_operand" "xAVpXD")]
+(define_insn "riscv_arcv_apex_<DM:mode>_ftype_v"
+  [(set (match_operand:DM 0 "register_operand" "=r")
+	(unspec_volatile:DM [(match_operand:SI 1 "const_int_operand" "xAVpXD")]
 	UNSPEC_ARCV_APEX_DEST_V))]
   ""
 {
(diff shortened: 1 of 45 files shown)


gcc/config/riscv/riscv-c.cc
```diff
@@ -100,6 +101,235 @@ riscv_pragma_intrinsic_flags_restore (struct pragma_intrinsic_flags *flags)
...
+  switch (token)
+  {
+    case CPP_STRING:
+      insn_name_raw = TREE_STRING_POINTER (x);
+      break;
+    case CPP_NAME:
+      insn_name_raw = IDENTIFIER_POINTER (x);
+      break;
+    default:
+      error ("pragma intrinsic: APEX attribute 'name' is missing");
```
MichielDerhaeg: Missing return?

gcc/config/riscv/riscv-c.cc
```diff
@@ -100,6 +101,235 @@ riscv_pragma_intrinsic_flags_restore (struct pragma_intrinsic_flags *flags)
...
+  /* Parse the function identifier to be marked as intrinsic.  */
+  if (pragma_lex (&x) != CPP_NAME)
+  {
+    error ("expected intrinsic name identifier");
+    return;
+  }
+  const char *fn_name = IDENTIFIER_POINTER (x);
+
+  /* Consume the comma.  */
+  pragma_lex (&x);
```
MichielDerhaeg, line 202: Check for comma, could be anything.

gcc/config/riscv/riscv-c.cc
```diff
@@ -100,6 +101,235 @@ riscv_pragma_intrinsic_flags_restore (struct pragma_intrinsic_flags *flags)
...
+      break;
+    case CPP_NAME:
+      insn_name_raw = IDENTIFIER_POINTER (x);
+      break;
+    default:
+      error ("pragma intrinsic: APEX attribute 'name' is missing");
+  }
+
+  /* Reject instruction names that do not follow APEX identifier rules.  */
+  if (insn_name_raw[0] != '\0'
```
MichielDerhaeg: And what if it is the empty string?

gcc/config/riscv/riscv-c.cc
```diff
@@ -100,6 +101,235 @@ riscv_pragma_intrinsic_flags_restore (struct pragma_intrinsic_flags *flags)
...
+  }
+
+  /* Convert instruction name to lowercase to normalize it
+     for the assembler.  */
+  char *insn_name = xstrdup (insn_name_raw);
+  for (char *p = insn_name; *p; p++)
+    *p = TOLOWER (*p);
+
+  /* Consume the comma.  */
+  pragma_lex (&x);
```
MichielDerhaeg, line 227: Idem.

MichielDerhaeg (changes requested):

gcc/config/riscv/arcv-apex.md
```diff
@@ -0,0 +1,437 @@
...
+  UNSPEC_ARCV_APEX_DEST_V
+  UNSPEC_ARCV_APEX_DEST_SRC0
+  UNSPEC_ARCV_APEX_DEST_SRC0_V
+  UNSPEC_ARCV_APEX_DEST_SRC0_SRC1
+  UNSPEC_ARCV_APEX_DEST_SRC0_SRC1_V
+])
+
+;; Used by "XD" insn. format: `insn`
+(define_insn "riscv_arcv_apex_void_ftype_v"
+  [(unspec_volatile:SI [(match_operand:SI 0 "const_int_operand" "xAVpXD")]
```
MichielDerhaeg, line 36: How is it determined whether this needs to be volatile? I know it is specified in the syntax, but how is the right insn picked?

gcc/config/riscv/riscv-builtins.cc
```diff
@@ -369,6 +473,17 @@ riscv_expand_builtin_direct (enum insn_code icode, rtx target, tree exp,
...
 
+  if (has_subcode_p)
+  {
+    /* Create an RTL constant for the APEX subcode.  */
+    rtx const_rtx = GEN_INT (subcode);
+    /* Add the subcode as an additional input operand to the RTL expression.  */
+    create_input_operand (&ops[opno++], const_rtx, SImode);
+    /* Validate the immediate argument passed to the APEX intrinsic.  */
+    if (!arcv_apex_immediate_argument_valid_p (subcode, exp))
+      return const0_rtx;
```
MichielDerhaeg, line 502: What happens if it isn't valid?

gcc/config/riscv/riscv-builtins.cc
```diff
@@ -420,16 +535,33 @@ riscv_expand_builtin (tree exp, rtx target, rtx subtarget ATTRIBUTE_UNUSED,
...
+    {
+    case RISCV_BUILTIN_DIRECT:
+      return riscv_expand_builtin_direct (d->icode, target, exp, true,
+					  subcode, true);
+
+    case RISCV_BUILTIN_DIRECT_NO_TARGET:
+      return riscv_expand_builtin_direct (d->icode, target, exp, false,
+					  subcode, true);
+    }
+    break;
```
MichielDerhaeg: Formatting looks a bit off here.

gcc/config/riscv/riscv-builtins.cc
```diff
@@ -454,3 +586,421 @@ riscv_atomic_assign_expand_fenv (tree *hold, tree *clear, tree *update)
...
+    /* We want to check the size of the value represented by the argument.
+       - If it's a pointer, we check the size of the pointed-to type.
+       - If it's a scalar or aggregate type, we check its own size.  */
+    tree argtype = TREE_VALUE (arg);
+    tree type_to_check = argtype;
+    if (TREE_CODE (argtype) == POINTER_TYPE)
+    {
+      /* Skip size check if return type is void.  */
+      if (return_type == void_type_node)
+	continue;
```
MichielDerhaeg, line 692: Why does the return type influence anything if the arg type is a pointer?

MichielDerhaeg: I just read this:
> APEX function %qs must return the same type as the first parameter for the %qs format class

But in this case they are not the same?

gcc/config/riscv/riscv-builtins.cc
```diff
@@ -454,3 +586,421 @@ riscv_atomic_assign_expand_fenv (tree *hold, tree *clear, tree *update)
...
+
+  /* Modify the prototype type as built-in.  */
+  fndecl->function_decl.built_in_class = BUILT_IN_MD;
+
+  /* Modify the prototype function code to match the index
+     in "riscv_apex_builtins" with a mask for APEX only insns.  */
+  fndecl->function_decl.function_code
+	= (arcv_apex_builtin_index << RISCV_BUILTIN_SHIFT) + RISCV_BUILTIN_APEX;
+
+  arcv_apex_builtin_index++;
```
MichielDerhaeg, line 1024: More like a size/amount than an index, no?

MichielDerhaeg (changes requested):

gcc/config/riscv/riscv-c.cc
luismgsilva: The intention was for the compiler to report all missing attributes without considering the presence of a comma. That’s how CCAC behaves.

For example:
 ```c
int foo(int);

#pragma intrinsic(foo) 
```
It complains with:
```bash
arcv-apex-err5.c:6:9: error: pragma intrinsic: APEX attribute 'name' is missing
    6 | #pragma intrinsic(foo)
      |         ^~~~~~~~~
arcv-apex-err5.c:6:9: error: pragma intrinsic: APEX attribute 'opcode' is missing
```

When we add extra validations for the comma, we end up with output like:
```bash
arcv-apex-err5.c:6:9: error: pragma intrinsic: expected ',' after APEX attribute name 'foo'
    6 | #pragma intrinsic(foo)
      |         ^~~~~~~~~
arcv-apex-err5.c:6:9: error: pragma intrinsic: APEX attribute 'name' is missing
arcv-apex-err5.c:6:9: error: pragma intrinsic: expected ',' after APEX attribute name ''
arcv-apex-err5.c:6:9: error: pragma intrinsic: APEX attribute 'opcode' is missing
```

If we’re okay with having this difference in behavior between the two compilers, I have no objection.

gcc/config/riscv/riscv-c.cc
luismgsilva: In some cases, we want the compiler to keep executing so it can report all errors related to the pragma syntax at once.

gcc/config/riscv/riscv-c.cc
luismgsilva: Good catch. Fixed.

gcc/config/riscv/arcv-apex.md
luismgsilva: If an instruction is defined with a void return type and no arguments, then that’s the only valid form it can take. See:
https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/blob/861b8c603897e88c16c563e55764cf82d18ff124/gcc/config/riscv/riscv-builtins.cc#L905

Any instruction without a return value (and therefore no destination register) is marked as volatile.

gcc/config/riscv/riscv-builtins.cc
luismgsilva: The validations happen inside the `arcv_apex_immediate_argument_valid_p` function. If an error is triggered, the caller simply exits by returning const0_rtx.

gcc/config/riscv/riscv-builtins.cc
luismgsilva: > Why does the return type influence anything if the arg type is a pointer?

If an intrinsic has a void return type, it can accommodate a larger size.

luismgsilva: > I just read this:
> 
> > APEX function %qs must return the same type as the first parameter for the %qs format class
> 
> But in this case they are not the same?

For "XC" instruction format the return type and first argument must be the same type. Because the destination register and first argument operand are the same register.

gcc/config/riscv/riscv-builtins.cc
luismgsilva: I understand your point regarding the naming. `arcv_apex_builtin_index` is used both as an index for inserting the next entry and as the current size for iteration bounds. If there’s a preference for clarity, I’m fine renaming it to `arcv_apex_builtin_count`. Otherwise, I kept index since its primary role is writing into the array.

gcc/config/riscv/riscv-builtins.cc
luismgsilva: I kept it close to the existing pattern in `riscv_expand_builtin` to minimize stylistic discrepancies

gcc/config/riscv/riscv-c.cc
MichielDerhaeg: It does look cleaner indeed if we omit the comma error.
But if we don't check for it, it will parse it wrong if you forget about the comma.

Can we check for it, but delay the error, and omit it when it makes sense?

gcc/config/riscv/riscv-builtins.cc
MichielDerhaeg: I was just nitpicking.

gcc/config/riscv/riscv-c.cc
MichielDerhaeg: And if you really don't want to check it, at least document why.

gcc/config/riscv/riscv-builtins.cc
MichielDerhaeg: Everything inside the `RISCV_BUILTIN_APEX` needs to be indented a bit more.

gcc/config/riscv/riscv-builtins.cc
```diff
@@ -454,3 +586,421 @@ riscv_atomic_assign_expand_fenv (tree *hold, tree *clear, tree *update)
...
+   instructions.
+
+   This function retrieves the instruction name string corresponding to the
+   specified subcode, allowing the backend code to reference the mnemonic
+   of the builtin instruction during assembly emission.  */
+
+const char*
+arcv_apex_get_insn_name (rtx op)
+{
+	unsigned int subcode = UINTVAL (op);
```
MichielDerhaeg: Formatting. Is now a tab, should be 2 spaces.

MichielDerhaeg (changes requested):

gcc/config/riscv/riscv-builtins.cc
luismgsilva: Fixed. Thanks

gcc/config/riscv/riscv-builtins.cc
luismgsilva: Fixed. Thanks.

gcc/config/riscv/riscv-c.cc
luismgsilva: Since omitting the comma always leads to a missing attribute error anyway, reporting a separate comma error would be redundant. We could technically validate the comma during parsing and only report it if no other attribute-related errors are found - though in practice, that situation doesn’t occur.
For clarity, I added a comment in the source code to document the reasoning.

MichielDerhaeg (approved):

luismgsilva:
The most recent force push addressed the comments in the emitted instructions. For RISC-V, the standard assembly comment symbol is `#`, whereas for ARC it is `;`.

PR #161, opened 2025-08-13
Fix XARCVvsad intrinsic arguments and return types

MichielDerhaeg:

### DIFF
--- a/gcc/config/riscv/riscv-vector-builtins-bases.cc
+++ b/gcc/config/riscv/riscv-vector-builtins-bases.cc
@@ -3289,7 +3289,8 @@ public:
     switch (e.op_info->op)
       {
       case OP_TYPE_vv:
-	return e.use_exact_insn (code_for_pred_widen_arcv_vwsad (e.vector_mode ()));
+	return e.use_widen_ternop_insn (
+	    code_for_pred_widen_arcv_vwsad (e.vector_mode ()));
       default:
 	gcc_unreachable ();
       }
@@ -3306,7 +3307,8 @@ public:
     switch (e.op_info->op)
       {
       case OP_TYPE_vv:
-	return e.use_exact_insn (code_for_pred_widen_arcv_vwsadu (e.vector_mode ()));
+	return e.use_widen_ternop_insn (
+	    code_for_pred_widen_arcv_vwsadu (e.vector_mode ()));
       default:
 	gcc_unreachable ();
       }
--- a/gcc/config/riscv/riscv-vector-builtins.cc
+++ b/gcc/config/riscv/riscv-vector-builtins.cc
@@ -1213,6 +1213,12 @@ static CONSTEXPR const rvv_arg_type_info v_sx_args[]
      rvv_arg_type_info (RVV_BASE_size),
      rvv_arg_type_info (RVV_BASE_vector),
      rvv_arg_type_info_end};
+static CONSTEXPR const rvv_arg_type_info ss_wwvv_args[]
+  = {rvv_arg_type_info (RVV_BASE_vector),
+     rvv_arg_type_info (RVV_BASE_double_trunc_signed_vector),
+     rvv_arg_type_info (RVV_BASE_double_trunc_signed_vector),
+     rvv_arg_type_info_end};
+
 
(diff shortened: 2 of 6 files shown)


gcc/config/riscv/riscv-vector-builtins.cc
```diff
@@ -893,6 +893,12 @@ static CONSTEXPR const rvv_arg_type_info su_wwvv_args[]
      rvv_arg_type_info (RVV_BASE_double_trunc_unsigned_vector),
      rvv_arg_type_info_end};
 
+static CONSTEXPR const rvv_arg_type_info vwsad_args[] // FIXME Name.
```
luismgsilva: "FIXME Name"?

gcc/testsuite/gcc.target/riscv/arcv-vsad-vwsad_vv-compile-1.c
```diff
@@ -5,13 +5,13 @@
...
+vuint16m2_t test_vwsad_vv_i8 (vuint16m2_t vd, vint8m1_t vs2, vint8m1_t vs1, size_t vl) {
+  return __riscv_arcv_vwsad_vv_u16m2 (vd, vs2, vs1, vl); }
+vuint16m2_t test_vwsad_vv_i8_m (vbool8_t mask, vuint16m2_t vd, vint8m1_t vs2, vint8m1_t vs1, size_t vl) {
+  return __riscv_arcv_vwsad_vv_u16m2_m (mask, vd, vs2, vs1, vl); }
+vuint32m2_t test_vwsad_vv_i16 (vuint32m2_t vd, vint16m1_t vs2, vint16m1_t vs1, size_t vl) {
+  return __riscv_arcv_vwsad_vv_u32m2 (vd, vs2, vs1, vl); }
+vuint32m2_t test_vwsad_vv_i16_m (vbool16_t mask, vuint32m2_t vd, vint16m1_t vs2, vint16m1_t vs1, size_t vl) {
+  return __riscv_arcv_vwsad_vv_u32m2_m (mask, vd, vs2, vs1, vl); }
 
-/* { dg-final { scan-assembler-times "arcv\\.vwsad\\.vv" 4 } } */
```
luismgsilva, line 17: Did you unintentionally removed a newline?

gcc/config/riscv/riscv-vector-builtins-bases.cc
```diff
@@ -3289,7 +3289,7 @@ class arcv_vwsad : public function_base
     switch (e.op_info->op)
       {
       case OP_TYPE_vv:
-	return e.use_exact_insn (code_for_pred_widen_arcv_vwsad (e.vector_mode ()));
+	return e.use_widen_ternop_insn (code_for_pred_widen_arcv_vwsad (e.vector_mode ()));
```
luismgsilva: I believe this line exceeds 80 characters. You can check for GNU style using:
```bash
$ git format-patch -1
$ bash contrib/check_GNU_style.sh 0001-Fix-XARCVvsad-intrinsic-arguments-and-return-types.patch

Lines should not exceed 80 characters.
54:+        return e.use_widen_ternop_insn (code_for_pred_widen_arcv_vwsad (e.vector_mode ()));
63:+        return e.use_widen_ternop_insn (code_for_pred_widen_arcv_vwsadu (e.vector_mode ()));

...
```
I also think that related style issues from the original commit may not have been addressed.  Not sure whether we should align all of them.

gcc/testsuite/gcc.target/riscv/arcv-vsad-vwsad_vv-compile-1.c
MichielDerhaeg: Other way around. The original DSP tests all have a missing newline at the end, which the editor adds in whenever you save the file. Something that should be fixed for the other tests, but I won't do it now.

gcc/config/riscv/riscv-vector-builtins-bases.cc
MichielDerhaeg: Won't fix the other ones here. We'll do it later. We will have to when we upstream.

luismgsilva (approved):
LGTM. Thanks!

PR #162, opened 2025-08-22
arcv: Add option to indicate the core supports advanced fusions

MichielDerhaeg:


MichielDerhaeg:
I'm going to merge it in both .06 and .09 if we want to to performance measurements with this. .09 still has known regressions.

Alexehv77 (approved):

gcc/testsuite/gcc.target/riscv/arcv-advanced-fusion-1.c
```diff
@@ -0,0 +1,12 @@
+/* { dg-do compile } */
+/* { dg-require-effective-target rv32 } */
+/* { dg-skip-if "" { *-*-* } { "-g" "-flto" "-O0" } } */
+/* { dg-options "-mtune=arc-v-rhx-100-series -mno-arc-v-rmx-500-series-advanced-fusion -march=rv32im -mabi=ilp32" } */
```
luismgsilva, line 4: If rmx500 does not support advanced fusion, why testing the enable/disable flag? Shouldnt we test for the other mtunes?

MichielDerhaeg: The point of the flag is to be able to turn off this fusion because the RMX-500 has a base version without it, and a perf version with the fusion. Always enabling the fusion degrades performance if the core doesn't support it.

The flag should only apply to the rmx 500 mtune, but because that mtune doesn't handle fusion yet, we can only test it with the rhx or rpx mtunes.

luismgsilva (approved):

PR #163, opened 2025-08-26
RISC-V: Add -mno-compress option

MichielDerhaeg:
This option prevents GCC from emitting compressed instructions even if the extensions are specified.


luismgsilva (approved):
LGTM. Thanks!

kolerov:
@MichielDerhaeg Could you please push it to `arc-2025.09`?

PR #164, opened 2025-08-29
arc64: Update symbol-2.c to use strict function prototypes.

luismgsilva:
Changed the test case to declare separate functions with explicit prototypes:

- `f1()` replaces the old K&R-style `f()` for calls with no arguments.
- `f2(int,int,int,void*)` is used when passing a pointer to `_nl_C_locobj`.
- `f3(int,int,int,struct s)` is used when passing `_nl_C_locobj` by value.

This ensures the test compiles cleanly on GCC 15, preserving the `_nl_C_locobj` access patterns for the assembly scan.

### DIFF
--- a/gcc/testsuite/gcc.target/arc64/symbol-2.c
+++ b/gcc/testsuite/gcc.target/arc64/symbol-2.c
@@ -7,7 +7,9 @@ struct s {
   int a;
 } _nl_C_locobj;
 
-extern int f();
+extern int f1();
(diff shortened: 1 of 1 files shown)


MichielDerhaeg:
Ah, it's because of this: https://gcc.gnu.org/gcc-15/porting_to.html#c23-fn-decls-without-parameters

MichielDerhaeg (approved):

PR #165, opened 2025-09-02
Use correct return type for ARCVvdsp vmv intrinsic

MichielDerhaeg:
- The result is placed in a scalar-in-vector register. The type of vd is
      always LMUL=1.
    - The naming of the intrinsic now includes type of vs2 first, and the
      type of vd second.


MichielDerhaeg:
I'm unsure about what the difference between `V_VLSI` and `V_VLSI:MODE` is. Not sure whether that needs to be changed.

luismgsilva (approved):

PR #167, opened 2025-09-12
Partially revert "Add helpers to test whether an optab can be implemented"

MichielDerhaeg:
Reverts the changes from 485ab50c204ddba922c92d8d467ac734428a4e8a in tree-vect-generic.cc and tree-vect-stmts.cc. This fixes the following failing tests for the arc & arc64 target:
 - gcc.dg/torture/pr52407.c
 - gcc.c-torture/execute/pr23135.c
 - gcc.c-torture/execute/pr53645.c
 - gcc.c-torture/execute/scal-to-vec1.c
 - gcc.c-torture/execute/scal-to-vec2.c
 - gcc.c-torture/execute/simd-1.c
 - gcc.dg/pr88598-4.c

The new helpers the reverted commit introduced are left in place because they're used in new places after they were comitted.

### DIFF
--- a/gcc/tree-vect-generic.cc
+++ b/gcc/tree-vect-generic.cc
@@ -1598,29 +1598,49 @@ ssa_uniform_vector_p (tree op)
   return NULL_TREE;
 }
 
-/* Return the type that should be used to implement OP on type TYPE.
-   This is TYPE itself if the target can do the operation directly,
-   otherwise it is a scalar type or a smaller vector type.  */
+/* Return type in which CODE operation with optab OP can be
+   computed.  */
 
 static tree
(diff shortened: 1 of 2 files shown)


luismgsilva:
@MichielDerhaeg Can you mention on the commit message which target/CPU those tests are failing? Or is it for all?
Gets easier once we come back to fix it.

MichielDerhaeg:
> @MichielDerhaeg Can you mention on the commit message which target/CPU those tests are failing? Or is it for all? Gets easier once we come back to fix it.

It's not specific to a specific arc target. But I'll mention arc(64) to ensure we won't look at risc-v.

luismgsilva (approved):

PR #169, opened 2025-09-13
Revert enabling -gvariable-location-views by default

kolerov:
This PR reverts a couple of patches in GCC 15 which turn on `-gvariable-location-views` by default (as it said in the original commit message). Somehow this behavior leads to significant increase of binaries' size. Mostly it's debug symbols and relocations for them. E.g., our toolchain with GCC 15 without this fix is 15GB, and with this fix it is 11GB.

In the future releases we should investigate how to overcome this issue without reverting of these patches.

### DIFF
--- a/gcc/dwarf2out.cc
+++ b/gcc/dwarf2out.cc
@@ -10372,10 +10372,8 @@ dwarf2out_maybe_output_loclist_view_pair (dw_loc_list_ref curr)
     }
   else
     {
-      dw2_asm_output_data_uleb128 (ZERO_VIEW_P (curr->vbegin)
-				   ? 0 : curr->vbegin, "Location view begin");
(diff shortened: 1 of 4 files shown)


MichielDerhaeg:
Wouldn't it be possible to simply build the toolchain with `-gno-variable-location-views` instead of changing defaults?

kolerov:
@MichielDerhaeg Sure, this might work. I will check it with Crosstool-NG builds.

kolerov:
@MichielDerhaeg I see that simply passing that `-gno-...` options does not reduce binaries size well. That reverted patch also change a way how debug symbols are generate even without that option. I suggest applying this PR for this release and we will try to figure out to get rid of this behavior in the next release or how to tune it properly.

MichielDerhaeg (approved):
Ok, this change isn't advertised in the GCC 15 release notes, so is fine I think to release with revert.

PR #170, opened 2025-09-30
Revert "RISC-V: Add -mno-compress option"

MichielDerhaeg:
This functionality is instead provided in the tcf script.


kolerov:
RC1 tags are already created, and I've made `arc-2025.09` branch read only. So, we should push it to RC2 (if it will be necessary) or to the next release branch.

kolerov (approved):

PR #174, opened 2025-10-07
arcv: Add rounding mode support for ARCVvdsp.

luismgsilva:
Adds rounding mode support to the ARCVvdsp VSRA family of shift instructions, complex multiply instructions and vsaddsub instruction.
The rounding mode is now passed as a parameter in the associated intrinsic.

Affected instructions:
- vsra, vsra_s, vsra_2s
- vnsra, vnsra_s, vnsra_2s
- vwsra
- vsaddsub
- vsmulf
- vscmul
- vscjmul

Corresponding test cases have been updated to cover this change.


gcc/testsuite/gcc.target/riscv/arcv-vcplx-vscjmul_vv-compile-1.c
```diff
@@ -1,17 +1,60 @@
 /* { dg-do compile } */
 /* { dg-require-effective-target arcv_vcplx } */
-/* { dg-options "-march=rv32im_xarcvvcplx -mabi=ilp32" } */
+/* { dg-skip-if "" { *-*-* } { "-O0" "-O1" "-O3" "-Os" "-Og" "-Oz" "-flto" } } */
+/* { dg-options "-march=rv32im_xarcvvcplx -mabi=ilp32 -O2" } */
```
MichielDerhaeg, line 4: I realise this comes from my PR, but I wouldn't both exclude -O2 in dg-skip-if and add it in in dg-options.

MichielDerhaeg (approved):
Otherwise LGTM

gcc/testsuite/gcc.target/riscv/arcv-vcplx-vscjmul_vv-compile-1.c
luismgsilva: So you think it should be as follows?
```
/* { dg-skip-if "" { *-*-* } { "-O0" "-O1" "-O2" "-O3" "-Os" "-Og" "-Oz" "-flto" } } */
/* { dg-options "-march=rv32im_xarcvvcplx -mabi=ilp32 -O2" } */
```

MichielDerhaeg: That won't work, you'll skip everything.
I'm just saying that skipping everything & specifying -O2 in dg-options is redundant.
Either we remove -O2 in dg-options, or there is a better option to specify the optimization level without having to run the test N times for every optimization flag.

luismgsilva: That's what I also thought.
It runs for those options because it's set on the: https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/blob/f83a33539af302e592b997e0759a07eef12ac69b/gcc/testsuite/gcc.target/riscv/riscv.exp#L45

For APEX, I created a separate .exp file to be able to run APEX tests seperately, and also remove those options: [apex.exp](https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/blob/arc-2025.09/gcc/testsuite/gcc.target/riscv/apex/apex.exp)

luismgsilva:
The https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/174/commits/aba576ff8f028afcd1f0689751602f59a0bf3aa4 has been included in this PR since it relies on the testsuite changes introduced by the earlier patches.

Fixes the `vnsra_s` and `vnsra_2s` intrinsics to correctly accept LMUL=1 for the first argument. Updates patterns, argument descriptors, builtins, and tests. Overload support is disabled due to conflicts between identical argument lists.

MichielDerhaeg (approved):
LGTM

PR #177, opened 2025-10-15
arc: Fix wrong vector ordering on big-endian architecture

exur00:
V2HI vectors, explicitly or auto-generated, could be stored in memory
wrongly due to endianness. This caused 2 values to be swapped in the
snprintf.c for big-endian hs38_linux cpu.

This patch swaps the 2 values if the target is big-endian.

### DIFF
--- a/gcc/config/arc/simdext.md
+++ b/gcc/config/arc/simdext.md
@@ -1438,11 +1438,12 @@
   "reload_completed && GET_CODE (operands[1]) == CONST_VECTOR"
   [(set (match_dup 0) (match_dup 2))]
   {
-   HOST_WIDE_INT intval = INTVAL (XVECEXP (operands[1], 0, 1)) << 16;
-   intval |= INTVAL (XVECEXP (operands[1], 0, 0)) & 0xFFFF;
-
-   operands[0] = gen_rtx_REG (SImode, REGNO (operands[0]));
-   operands[2] = GEN_INT (trunc_int_for_mode (intval, SImode));
+    int hi = !TARGET_BIG_ENDIAN;
+    int lo = !hi;
+    HOST_WIDE_INT intval = INTVAL (XVECEXP (operands[1], 0, hi)) << 16;
+    intval |= INTVAL (XVECEXP (operands[1], 0, lo)) & 0xFFFF;
+    operands[0] = gen_rtx_REG (SImode, REGNO (operands[0]));
+    operands[2] = GEN_INT (trunc_int_for_mode (intval, SImode));
   }
   [(set_attr "type" "move,move,load,store")
    (set_attr "predicable" "yes,yes,no,no")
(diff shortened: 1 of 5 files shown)


gcc/config/arc/simdext.md
```diff
@@ -1437,12 +1437,19 @@
...
+    if(TARGET_BIG_ENDIAN)
+      {
+        HOST_WIDE_INT intval = INTVAL (XVECEXP (operands[1], 0, 0)) << 16;
+        intval |= INTVAL (XVECEXP (operands[1], 0, 1)) & 0xFFFF;
+      }
+    else
+      {
+        HOST_WIDE_INT intval = INTVAL (XVECEXP (operands[1], 0, 1)) << 16;
+        intval |= INTVAL (XVECEXP (operands[1], 0, 0)) & 0xFFFF;
+      }
```
MichielDerhaeg: Don't copy the entire statements in the if & else branch. If only the indices need to change only assign those.
Similar to mulsidi_600 in arc.md.

gcc/testsuite/gcc.target/arc/big_endian_subregister_vectorisation.c
```diff
@@ -0,0 +1,16 @@
...
+
+void bar()
+{
+    v2hi v;
+    v[0] = 520;
+    v[1] = -1;
+    foo(&v);
+}
+
+/* { dg-final { scan-assembler-not "-65016" } } */
```
MichielDerhaeg: I would also check the instruction mnemonic, not just the value of the immediate.

MichielDerhaeg (changes requested):

gcc/testsuite/gcc.target/arc/movv2hi-be.c
MichielDerhaeg, line 1: Can you give this test a less verbose name? Something like movv2hi-be?

MichielDerhaeg (changes requested):
I would elaborate a bit more in the commit message which issues you're fixing. Also mention this breaks stores of struct members if these operations are SLP vectorized, causing them to be stored in the wrong order. You can be very verbose in your commit message.
  - put in the minimal c test that gets vectorized
    - the struct type on 1 line, and the two assignments is sufficient (so just 3 lines of c code)
  - show the RTL before the split pass (just the mov insn)
  - show the RTL after the split pass and say this is obviously wrong for BE (idem, just the mov insn)

> This caused 2 values to be swapped in the snprintf.c for big-endian hs38_linux cpu.
This can be dropped from the commit message. Doesn't really contribute anything where this issue was discovered.

gcc/testsuite/gcc.target/arc/movv2hi-be.c
```diff
@@ -0,0 +1,32 @@
...
+
+int main (void)
+{
+    v2hi v;
+    v[0] = 520;
+    v[1] = -1;
+    foo(&v);
+    foo2(v);
+    return 0;
+}
```
luismgsilva: The file should have a new line at the end.

luismgsilva (approved):
Great work! Thanks.
Just a note - try to keep each line in the commit message description under 80 characters, as longer lines can be harder to read.

MichielDerhaeg (approved):
regtests look fine, good to merge

PR #178, opened 2025-10-17
arcv: Fix ARCVvdsp vnsra_{2,}s intrinsics to use vint*m1_t as first arg.

luismgsilva:
The vnsra_s and vnsra_2s intrinsics were incorrectly defined to take
    vint*m4_t vectors as their first input argument, while their semantics
    require vint*m1_t inputs corresponding to LMUL=1 narrow results.

    This patch updates the affected operand types and patterns to use
    <V_LMUL1> in arcv-vector.md, and adjusts argument type descriptors
    in riscv-vector-builtins.cc accordingly.  The tests for vnsra_{2,}s
    intrinsics have been corrected to match the proper vint*m1_t argument
    types.

    After updating the argument type to use LMUL=1, multiple intrinsics
    derived from the same vnsra_{2,}s instruction triggered overload conflicts.
    The RVV overload mechanism only compares argument types (ignoring return
    types), so identical argument lists across these intrinsics caused
    the compiler to complain.  To address this, RVV overload support for
    vnsra was disabled.


luismgsilva:
This patch has been added to the PR https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/174  as it depends on the changes for the test cases.

PR #180, opened 2025-10-17
arcv: Fix ARCVvdsp vqmxm16su operand from unsigned to signed.

luismgsilva:
The vqmxm16su intrinsic was incorrectly defined with an unsigned operand (using su_qqvv_ops), but the instruction expects both operands to be signed.

This patch introduces su_qvv_ops which correctly specifies both narrow operands as signed quad-truncated vectors, and updates the vqmxm16su builtin definition to use it.  The test case has been corrected to use vint8m1_t instead of vuint8m1_t.


luismgsilva:
Current implementation is correct. The issue in the other compiler.

PR #181, opened 2025-10-20
arc: emit clobber of CC for -mcpu=em x >> 31

exur00:
Address the issue explained here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=120375

If the target has no barrel shifter, clobber is now emitted in a parallel RTL expression. si3_nobs matches the parallel expression and emits the necessary expressions that may or may not clobber the CC. This happens in split1, before registers are assigned and dead stores are removed. If the clobber was unnecessary, its consequences will be removed by those steps.

### DIFF
--- a/gcc/config/arc/arc.md
+++ b/gcc/config/arc/arc.md
@@ -3554,7 +3554,14 @@ archs4x, archs4xd"
   [(set (match_operand:SI 0 "dest_reg_operand" "")
 	(ANY_SHIFT_ROTATE:SI (match_operand:SI 1 "register_operand" "")
 			     (match_operand:SI 2 "nonmemory_operand" "")))]
-  "")
+  ""
+{
+  if (!TARGET_BARREL_SHIFTER && operands[2] != const1_rtx)
+    {
+      emit_insn (gen_<insn>si3_loop (operands[0], operands[1], operands[2]));
+      DONE;
+    }
+})
 
 ; asl, asr, lsr patterns:
 ; There is no point in including an 'I' alternative since only the lowest 5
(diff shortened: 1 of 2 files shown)


luismgsilva (approved):
Nothing to point out. Good Job! Thanks.

gcc/config/arc/arc.md
```diff
@@ -3653,35 +3660,27 @@ archs4x, archs4xd"
...
+  (clobber (reg:SI LP_COUNT))
+  (clobber (reg:CC CC_REG))]
+  "!TARGET_BARREL_SHIFTER"
   "* return output_shift_loop (<CODE>, operands);"
-  [(set_attr "type" "shift")
-   (set_attr "length" "16,20")])
+  "&& arc_pre_reload_split ()"
+  [(set (match_operand:SI 0 "dest_reg_operand" "=r,r")
+  (ANY_SHIFT_ROTATE:SI (match_operand:SI 1 "register_operand" "0,0")
+        (match_operand:SI 2 "nonmemory_operand" "rn,Cal")))]
```
MichielDerhaeg: use `match_dup` here, not `match_operand`

gcc/config/arc/arc.md
```diff
@@ -3653,35 +3660,27 @@ archs4x, archs4xd"
...
+{
+  if (operands[2] != const1_rtx)
+  {
+    arc_split_<insn> (operands);
+    DONE;
+  }
```
MichielDerhaeg: code-style requires extra indent of the braces

gcc/config/arc/arc.md
```diff
@@ -3554,7 +3554,14 @@ archs4x, archs4xd"
...
 	(ANY_SHIFT_ROTATE:SI (match_operand:SI 1 "register_operand" "")
 			     (match_operand:SI 2 "nonmemory_operand" "")))]
-  "")
+  ""
+{
+  if (!TARGET_BARREL_SHIFTER) && operands[2] != const1_rtx)
+  {
+    emit_insn (gen_<insn>si3_loop (operands[0], operands[1], operands[2]));
+    DONE;
+  }
```
MichielDerhaeg: code style

MichielDerhaeg (changes requested):

MichielDerhaeg (approved):

PR #182, opened 2025-10-21
arcv: Constrain vclr source operand to match destination.

luismgsilva:
The vclr instruction from ARCVvdsp operates in-place on vd,
clearing the lower imm5 elements while preserving the rest.
Fix the RTL pattern to constrain the source operand (operand 3)
to match the destination (operand 0).

Also update the assembly output template to the correct format
"arcv.vclr.v.i vd, imm5" without explicitly showing the source
register, and update test validations accordingly.

Note: the current intrinsic form may change in the future,
as vclr does not depend on LMUL or VL.


MichielDerhaeg (approved):
If this works, let's merge it like this.
But it seems like we're not sure what the intrinsic should look like given that it doesn't care about lmul and vl. This will change in the future.

Might be useful to mention this in the commit msg.

PR #183, opened 2025-10-27
arcv: Don't make xarcv imply v

MichielDerhaeg:
'v' implies 'zvl128b' and 'zve64d' while ARC-V hardware might be
configured to support smaller VLEN's or no FPU support.

- xarcvbit(rev|stream) are not RVV intrinsics
- other xarcv extensions now imply 'zve32x', the "smallest" RVV extension


luismgsilva (approved):
Looks good to me. Thanks!

PR #184, opened 2025-10-27
arcv: apex: Add LTO support for APEX intrinsics.

luismgsilva:
This patch creates a new lto section that serializes all the APEX intrinsics information
registered during GCC run-time.  Once `lto1` is invoked, this section is read and the
APEX intrinsics are re-registered.

A simple approach could be as follow:
1. Before looking into the GIMPLE representation tree, we store the intrinsic information
in the backend and generate a dummy function declaration.
2. When lto1 is checking in the backend for the availability of the intrinsic
(using the function code) it would find the dummy declaration.
3. Only after this validation GCC creates the actual `fndecl` from the GIMPLE representation
to be used for the expand.

This approach works for single compilation unit.  If we have multiple compilation units,
each with different APEX intrinsics (or same ones, but different registered order) the
`fncode` of each `fndecl` will overlap.

Example:
- Imagine that we have a `foo` in `file1.c` and `bar` in `file2.c`.  In each `fndecl` the
`fncode` would be `0`.
- When these files are used in `lto1` and we register a dummy intrinsic for each, the
intrinsic validation would use the same `fndecl` to validate both - we are not able to
distinguish which one is which from the `fncode`.

This patch's approach:
- We have to trick GCC to think that the intrinsic is available before its actually registered.
- It will then create the `fndecl` from the GIMPLE representation.
- The intrinsic registration is only handled after the GIMPLE representation and only after
all `fndecl` have been restored.
- This way, we can safely search for the `fndecl` of the given APEX intrinsic and change its
`fncode` to match what we are storing in the backend.

Each APEX intrinsic `fncode` (which is used as index in the `arcv_apex_builtins[]`) will have
the correct corresponding `fndecl` associated with it.

**Other comments**:
The read and write routines were initially in `gcc/` folder among other `ipa-*` files.
Since this implementation is riscv specific, I've decided to move it to the backend files.

**Testing**:
We are currently using a single `.exp` file to run both regular tests and lto tests. This can
be sperated in two `.exp` files (e.g., `apex.exp` and `apex-lto.exp`) if we decide so.
~~Validating lto error messages does not work as expected.  Patch removed from this PR.~~ - Fixed.

### DIFF
--- a/gcc/config/riscv/riscv-apex-lto.cc
+++ b/gcc/config/riscv/riscv-apex-lto.cc
@@ -1,0 +1,254 @@
+/* LTO serialization for RISC-V APEX intrinsics.
+   Copyright (C) 2025 Free Software Foundation, Inc.
+
+This file is part of GCC.
+
+GCC is free software; you can redistribute it and/or modify it under
+the terms of the GNU General Public License as published by the Free
+Software Foundation; either version 3, or (at your option) any later
+version.
+
+GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or
+FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+for more details.
+
+You should have received a copy of the GNU General Public License
+along with GCC; see the file COPYING3.  If not see
+<http://www.gnu.org/licenses/>.  */
+
+/* RISC-V APEX (ARC Processor Extension) intrinsics are unique in GCC
+   because they are registered dynamically at compile-time via #pragma intrinsic
+   directives, rather than being statically defined like normal target builtins.
+
+   This creates a challenge for LTO (Link Time Optimization): when compiling
+   with -flto, each translation unit may register different APEX intrinsics via
+   pragmas.  During the link-time optimization phase, all these intrinsic
+   definitions must be preserved and made available for code generation.
+
+   This file implements LTO serialization support for APEX intrinsics by:
+
+   1. Writing Phase (produce_asm_for_decls):
+      - Iterates through all registered APEX intrinsics
+      - Serializes their metadata (name, mnemonic, opcode, instruction formats)
+      - Writes to a dedicated .gnu.lto_riscv_apex section in object files
+
+   2. Reading Phase (read_cgraph_and_symbols):
+      - Reads .gnu.lto_riscv_apex sections from all input object files
+      - Reconstructs and re-registers all APEX intrinsics
+      - Makes them available for optimization and code generation
+
+   Without this support, LTO would lose APEX intrinsic definitions, causing
+   unavailable intrinsics errors during link-time optimization.  */
+
+#include "config.h"
+#include "system.h"
+#include "coretypes.h"
+#include "backend.h"
+#include "tree.h"
+#include "gimple.h"
+#include "cgraph.h"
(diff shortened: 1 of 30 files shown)


gcc/config/riscv/riscv-apex-lto.cc
```diff
@@ -0,0 +1,231 @@
...
+      insn_name[insn_name_len] = '\0';
+
+      /* Read opcode and instruction format flags.  */
+      unsigned int opcode = streamer_read_uhwi (ib);
+      unsigned int insn_formats = streamer_read_uhwi (ib);
+
+      /* Look up the function declaration in the merged symbol table.
+	 During LTO, all function declarations from all compilation units
+	 are merged into a single global symbol table.  */
+      symtab_node *snode = symtab_node::get_for_asmname (
```
MichielDerhaeg, line 222: What if the function is not found? Could it be optimized away if unused?

gcc/config/riscv/riscv-builtins.cc
```diff
@@ -373,6 +373,12 @@ riscv_builtin_decl (unsigned int code, bool initialize_p ATTRIBUTE_UNUSED)
 
     case RISCV_BUILTIN_VECTOR:
       return riscv_vector::builtin_decl (subcode, initialize_p);
+
+    case RISCV_BUILTIN_APEX:
+      /* Trick GCC to think that the function is defined.
+	 The actual fndecl used is created after this
+	 validation from the GIMPLE representation in LTO.  */
+      return integer_zero_node;
```
MichielDerhaeg, line 381: I won't let this prevent you from merging it now. But this is probably not a good long-term solution. Should we ask the community what a better solution could be?

gcc/config/riscv/riscv-apex-lto.cc
luismgsilva: If the function is not found, registration won’t occur, since I ensure that `node` is available in [here](https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/184/files/bf5c0bfd6722109f8c3bb4b913ec49b0341a918c#diff-9643ada2a065a9f678228ee3e80e6858ece90924632980b19f534375ce6e7a3fR203)

That said, an optimized-away apex intrinsic will still trigger the `gcc_assert (registered_count == apex_count);` assertion, which isn’t desirable.

We can remove the assertion, though ideally, we should only emit those that aren’t optimized away.
I will look into it.

gcc/config/riscv/riscv-builtins.cc
luismgsilva: Indeed, I also don't like this approach, and I doubt it would be accepted upstream. Although, I don't see a proper solution to identify the correct intrinsic by only the function code. Because with multiple compilation units, these fncodes might overlap.

gcc/config/riscv/riscv-apex-lto.cc
luismgsilva: @MichielDerhaeg Actually, intrinsics are not optimized away at tree level because they might have side-effects.  However, at RTL, it might be optimized away.
The issue that I describe only happens if the apex intrinsic is defined but never used.

I've made changes to only serialize apex intrinsics that are actually called. It might still be safer to keep this change.

MichielDerhaeg (approved):
LGTM

PR #186, opened 2025-10-28
arc: emit clobber of CC for -mcpu=em x >> 31

exur00:
Fixed solution that was merged in https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/181 but was reverted because it caused issues. It has been built and tested on the jenkins without issues.

### DIFF
--- a/gcc/config/arc/arc.md
+++ b/gcc/config/arc/arc.md
@@ -3554,7 +3554,14 @@ archs4x, archs4xd"
   [(set (match_operand:SI 0 "dest_reg_operand" "")
 	(ANY_SHIFT_ROTATE:SI (match_operand:SI 1 "register_operand" "")
 			     (match_operand:SI 2 "nonmemory_operand" "")))]
-  "")
+  ""
+{
+  if (!TARGET_BARREL_SHIFTER && operands[2] != const1_rtx)
+    {
+      emit_insn (gen_<insn>si3_loop (operands[0], operands[1], operands[2]));
+      DONE;
+    }
+})
 
 ; asl, asr, lsr patterns:
 ; There is no point in including an 'I' alternative since only the lowest 5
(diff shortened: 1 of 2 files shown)


MichielDerhaeg (changes requested):
Implementation looks fine. Let's improve the commit message a bit. Help the reviewer understand what you did.

> If the target has no barrel shifter, clobber is now emitted in a parallel RTL expression. <insn>si3_loop matches the parallel expression and emits the necessary expressions that may or may not clobber the CC. This happens in split1, before registers are assigned and dead stores are removed. If the clobber was unnecessary, its consequences will be removed by those steps.

A bit too verbose IMO. Just describe the problem we're solving:

> Devices without a barrel shifter end up using a sequence of
> instructions. These can use the condition codes and/or loop count
> register, so those need to be marked as 'clobbered'. These clobbers were previously added only after split1, which is too late. This patch adds these clobbers from the beginning,  in the `define_expand`.

Explain succinctly the changes you did, why did you merge _nobs and _loop?

> <insn>si3_cnt1_clobber is placed below all the specific si3_cnt1 patterns.
It matches the <insn>si3_loop pattern in the case operand2 has become
const_int 1, and remove the clobber so the specific si3_cnt1 patterns
can be matched.
> If <insn>si3_cnt1_clobber is placed above these patterns, some si3_cnt
would try to emit extra clobbers to match it instead of their more
specific pattern that match exactly. This would cause clobbers of hard
registers, causing failure.

This is a bit hard to parse. But I don't think you need to mention this. I think the comment you wrote in the code is sufficient.

luismgsilva:
> Implementation looks fine. Let's improve the commit message a bit. Help the reviewer understand what you did.
> 
> > If the target has no barrel shifter, clobber is now emitted in a parallel RTL expression. si3_loop matches the parallel expression and emits the necessary expressions that may or may not clobber the CC. This happens in split1, before registers are assigned and dead stores are removed. If the clobber was unnecessary, its consequences will be removed by those steps.
> 
> A bit too verbose IMO. Just describe the problem we're solving:
> 
> > Devices without a barrel shifter end up using a sequence of
> > instructions. These can use the condition codes and/or loop count
> > register, so those need to be marked as 'clobbered'. These clobbers were previously added only after split1, which is too late. This patch adds these clobbers from the beginning,  in the `define_expand`.
> 
> Explain succinctly the changes you did, why did you merge _nobs and _loop?
> 
> > si3_cnt1_clobber is placed below all the specific si3_cnt1 patterns.
> > It matches the si3_loop pattern in the case operand2 has become
> > const_int 1, and remove the clobber so the specific si3_cnt1 patterns
> > can be matched.
> > If si3_cnt1_clobber is placed above these patterns, some si3_cnt
> > would try to emit extra clobbers to match it instead of their more
> > specific pattern that match exactly. This would cause clobbers of hard
> > registers, causing failure.
> 
> This is a bit hard to parse. But I don't think you need to mention this. I think the comment you wrote in the code is sufficient.

This, and perhaps a before and after assembly example might help.

exur00:
I rewrote the commit with a better explanation and some RTL and assembly from the test. I also added keithp (who reported the bug) as co-author.

MichielDerhaeg (approved):

luismgsilva (approved):
LGTM. Great commit message.

PR #187, opened 2025-10-29
fusion support for rmx500

Alexehv77:
added fusion support for rmx500

### DIFF
--- a/gcc/config/riscv/riscv.cc
+++ b/gcc/config/riscv/riscv.cc
@@ -861,6 +861,7 @@ bool
 arcv_micro_arch_supports_fusion_p (void)
 {
   return (riscv_is_micro_arch (arcv_rhx100)
+	  || riscv_is_micro_arch (arcv_rmx500)
 	  || riscv_is_micro_arch (arcv_rpx100));
 }
 
@@ -11475,7 +11476,8 @@ riscv_override_options_internal (struct gcc_options *opts)
 
   /* Enable advanced fusion for the ARC-V cores that support it unless
      explicitly disabled.
-     TODO Add arcv_rmx500 once it supports fusion.  */
+     advanced fusion for rmx500 is turned on by an option.
+   */
   if ((riscv_microarchitecture == arcv_rhx100
   	|| riscv_microarchitecture == arcv_rpx100)
       && (target_flags_explicit & MASK_ARCV_ADVANCED_FUSION) == 0)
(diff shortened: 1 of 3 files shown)


gcc/config/riscv/riscv.cc
```diff
@@ -11470,10 +11471,14 @@ riscv_override_options_internal (struct gcc_options *opts)
...
     }
 
+  if (opts->x_riscv_mno_compress)
+    opts->x_riscv_zc_subext = 0;
+
   /* Enable advanced fusion for the ARC-V cores that support it unless
      explicitly disabled.
      TODO Add arcv_rmx500 once it supports fusion.  */
   if ((riscv_microarchitecture == arcv_rhx100
+        || riscv_microarchitecture == arcv_rmx500
```
MichielDerhaeg:
```suggestion
    explicitly disabled. */
if ((riscv_microarchitecture == arcv_rhx100 
```
It was decided that the mtune for rmx500 shouldn't imply the advanced fusions. They have to be explicitly turned on with the option.

gcc/config/riscv/riscv.md
```diff
@@ -4948,6 +4948,7 @@
 (include "corev.md")
 (include "xiangshan.md")
 (include "arcv-rmx100.md")
+;;(include "arcv-rmx500_nf.md")
```
MichielDerhaeg: Leftover?

MichielDerhaeg (changes requested):
I get slightly worse performance on rmx500_perf using this patch as compared to using the rhx100 mtune. Do you know why? Can you look into it?

Alexehv77:
> I get slightly worse performance on rmx500_perf using this patch as compared to using the rhx100 mtune. Do you know why? Can you look into it?

I looked into it and it has to do with the change in the number of iterations inside the testing environment. This implies that there is a runtime behavior. For 100 iterations there improvements were slightly better for rmx500. But when the number of iterations went to 150 the rhx100 got slightly better.

gcc/config/riscv/riscv.md
Alexehv77: I removed it

gcc/config/riscv/riscv.cc
Alexehv77: I have let the advance fusions to take place only under the explicit flag

MichielDerhaeg (approved):
Sorry for the delay.

luismgsilva (approved):

PR #188, opened 2025-11-14
RHX Upstream patches

luismgsilva:


gcc/config/riscv/arcv-rhx100.md
```diff
@@ -0,0 +1,113 @@
...
+;(final_presence_set "arcv_rhx100_issueA_fuse1" "arcv_rhx100_issueA_fuse0")
+;(final_presence_set "arcv_rhx100_issueB_fuse1" "arcv_rhx100_issueB_fuse0")
+;(final_presence_set "arcv_rhx100_ALU_A_fuse1_early" "arcv_rhx100_ALU_A_fuse0_early")
+;(final_presence_set "arcv_rhx100_ALU_B_fuse1_early" "arcv_rhx100_ALU_B_fuse0_early")
+
+;; Bypasses
+;(define_bypass 0 "arcv_rhx100_alu_early_arith" "arcv_rhx100_store_insn" "riscv_store_data_bypass_p")
+(define_bypass 1 "arcv_rhx100_alu_early_arith" "arcv_rhx100_store_insn" "riscv_store_data_bypass_p")
+
+;(define_bypass 0 "arcv_rhx100_load_insn" "arcv_rhx100_store_insn" "riscv_store_data_bypass_p")
```
MichielDerhaeg: Let's not keep commented out code around.

gcc/config/riscv/riscv-c.cc
```diff
@@ -149,6 +149,9 @@ riscv_cpu_cpp_builtins (cpp_reader *pfile)
     builtin_define_with_int_value ("__riscv_th_v_intrinsic",
 				   riscv_ext_version_value (0, 11));
 
+  if (TARGET_ARCV_RHX100)
+    builtin_define ("__riscv_rhx");
+
```
MichielDerhaeg: I wonder why this was added, none of the other mtunes do this.
I don't see it used anywhere or tested anywhere by either compiler. Maybe let's leave it out.

gcc/config/riscv/arcv-rhx100.md
```diff
@@ -0,0 +1,113 @@
+;; DFA scheduling description of the Synopsys RHX-100 cpu
+;; for GNU C compiler
+;; Copyright (C) 2023 Free Software Foundation, Inc.
```
MichielDerhaeg: Update copyright years.

gcc/config/riscv/arcv.cc
```diff
@@ -0,0 +1,833 @@
+/* ARCV-specific macro-op fusion for RISC-V.
+   Copyright (C) 2025 Free Software Foundation, Inc.
```
MichielDerhaeg: Would describe this more general. "ARC-V-specific functions something something"

gcc/config/riscv/arcv.cc
```diff
@@ -0,0 +1,833 @@
...
+
+
+/* Return true if PREV and CURR should be kept together during scheduling.  */
+
+bool
+arcv_macro_fusion_pair_p (rtx_insn *prev, rtx_insn *curr)
+{
+  /* Never create sched groups with more than 2 members.  */
+  if (SCHED_GROUP_P (prev))
+    return false;
```
MichielDerhaeg: This is already done in the caller function.

gcc/config/riscv/arcv.cc
```diff
@@ -0,0 +1,833 @@
...
+
+      /* Check if multiply result is used in either operand of the addition.  */
+      if (REG_P (XEXP (curr_plus, 0))
+	 && REGNO (XEXP (curr_plus, 0)) == mult_dest_regno)
+       return true;
+
+      if (REG_P (XEXP (curr_plus, 1))
+	 && REGNO (XEXP (curr_plus, 1)) == mult_dest_regno)
+       return true;
+    }
```
MichielDerhaeg, line 398: The caller adds debug logs to each of these, I think we should also do that.

gcc/config/riscv/riscv.cc
```diff
@@ -10787,92 +10665,16 @@ riscv_sched_variable_issue (FILE *, int, rtx_insn *insn, int more)
...
-
-  return riscv_issue_rate ();
-}
-
-
 /* Implement TARGET_SCHED_MACRO_FUSION_P.  Return true if target supports
    instruction fusion of some sort.  */
 
-static bool
+bool
```
MichielDerhaeg: Why?

luismgsilva: Not sure. Introduced in https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/commit/ec76bcca4784
Removed.

gcc/config/riscv/arcv.cc
luismgsilva: I've updated the headers of `arcv.{cc,h}` to match the style used in other architecture-specific files. What do you think?

gcc/config/riscv/arcv.cc
luismgsilva: I've added the respective debug logs.

gcc/config/riscv/arcv.cc
```diff
@@ -0,0 +1,881 @@
...
+    if (!sched_state.pipeB_scheduled_p
+	&& (get_attr_type (ready[*n_readyp - 1]) == TYPE_LOAD
+	    || get_attr_type (ready[*n_readyp - 1]) == TYPE_STORE))
+    {
+      sched_state.alu_pipe_scheduled_p = sched_state.pipeB_scheduled_p = 1;
+      sched_state.cached_can_issue_more = 1;
+      return 1;
+    }
+    else if (get_attr_type (ready[*n_readyp - 1]) != TYPE_LOAD
+	|| get_attr_type (ready[*n_readyp - 1]) != TYPE_STORE)
```
MichielDerhaeg: Don't think this condition is correct, it is always true.

luismgsilva: Im wondering if this is what is supposed to be:
```
From 4f60a816bd1f2dd28e471b473460321e48e2a99c Mon Sep 17 00:00:00 2001
From: Luis Silva <luiss@synopsys.com>
Date: Tue, 18 Nov 2025 15:12:07 +0000
Subject: [PATCH] tmp: arcv: Fix incorrect condition and logic in reorder
 scheduler.

Signed-off-by: Luis Silva <luiss@synopsys.com>
---
 gcc/config/riscv/arcv.cc | 19 ++++++++++---------
 1 file changed, 10 insertions(+), 9 deletions(-)

diff --git a/gcc/config/riscv/arcv.cc b/gcc/config/riscv/arcv.cc
index 8354d838c72..f0a39a23cb7 100644
--- a/gcc/config/riscv/arcv.cc
+++ b/gcc/config/riscv/arcv.cc
@@ -670,22 +670,23 @@ arcv_sched_reorder2 (rtx_insn **ready, int *n_readyp)
   /* If all else fails, schedule a single instruction.  */
   if (ready && *n_readyp > 0
       && NONDEBUG_INSN_P (ready[*n_readyp - 1])
-      && recog_memoized (ready[*n_readyp - 1]) >= 0
-      && get_attr_type (ready[*n_readyp - 1]) != TYPE_LOAD
-      && get_attr_type (ready[*n_readyp - 1]) != TYPE_STORE)
+      && recog_memoized (ready[*n_readyp - 1]) >= 0)
   {
+    rtx_insn *insn = ready[*n_readyp - 1];
+    enum attr_type insn_type = get_attr_type (insn);
+
+    /* Memory operations go to pipeB if available.  */
     if (!sched_state.pipeB_scheduled_p
-       && (get_attr_type (ready[*n_readyp - 1]) == TYPE_LOAD
-           || get_attr_type (ready[*n_readyp - 1]) == TYPE_STORE))
+       && (insn_type == TYPE_LOAD || insn_type == TYPE_STORE))
     {
-      sched_state.alu_pipe_scheduled_p = sched_state.pipeB_scheduled_p = 1;
+      sched_state.pipeB_scheduled_p = 1;
       sched_state.cached_can_issue_more = 1;
       return 1;
     }
-    else if (get_attr_type (ready[*n_readyp - 1]) != TYPE_LOAD
-       || get_attr_type (ready[*n_readyp - 1]) != TYPE_STORE)
+    /* Non-memory operations go to ALU pipe.  */
+    else if (insn_type != TYPE_LOAD && insn_type != TYPE_STORE)
     {
-      sched_state.alu_pipe_scheduled_p = sched_state.pipeB_scheduled_p = 1;
+      sched_state.alu_pipe_scheduled_p = 1;
       sched_state.cached_can_issue_more = 1;
       return 1;
     }
--
2.39.5

```

MichielDerhaeg: I think this makes sense. Can we evaluate the impact of this change on performance downstream & upstream?

luismgsilva: I got the following between unpatched and patched for downstream and upstream:

**Downstream**
  Unpatched vs patched = +-0.15% faster

**Upstream**
Unpatched vs Patched = +-0.27% faster

MichielDerhaeg (approved):
Looks good. I think we can start splitting this up in a patch series for review.

PR #189, opened 2025-11-20
arc: check if the addend fits when referencing small data memory

MichielDerhaeg:


gcc/config/arc/arc.cc
```diff
@@ -386,27 +386,35 @@ legitimate_small_data_address_p (rtx x, machine_mode mode)
...
 	  case 1:
-	    return p0;
+	    break;
 	  case 2:
-	    return p0 && ((INTVAL (XEXP (x, 1)) & 0x1) == 0);
+	    if ((offset & 0x1) == 0) break; else return false;
 	  case 4:
-	  case 8:
-	    return p0 && ((INTVAL (XEXP (x, 1)) & 0x3) == 0);
+	    if ((offset & 0x3) == 0) break; else return false;
```
exur00, line 410: if size is 8, then `size == 8 ? 4 : size;` is 4, making 8 unreachable I would think

MichielDerhaeg: That's why the 8 case is removed. 8 should always be treated the same as 4.

exur00: You're right, I misread the diff

exur00 (approved):

kolerov (approved):

PR #190, opened 2025-11-27
Alex/rmx500 fusion p10019563 81217

Alexehv77:


gcc/config/riscv/riscv.cc
```diff
@@ -10859,7 +10859,7 @@ riscv_sched_reorder2 (FILE *file ATTRIBUTE_UNUSED,
 	  if (NONDEBUG_INSN_P (ready[*n_readyp - i])
 	      && !SCHED_GROUP_P (ready[*n_readyp - i])
 	      && (!next_insn (ready[*n_readyp - i])
-		  || !NONDEBUG_INSN_P (next_insn (ready[*n_readyp - i]))
+		 // || !NONDEBUG_INSN_P (next_insn (ready[*n_readyp - i]))
```
MichielDerhaeg: Don't just comment these out. Remove the lines if they need to be removed.
Also, don't we need something like `next_non_dbg_insn`? That keeps invoking `next_insn` in a loop until it finds either a non-debug insn or null?

MichielDerhaeg (changes requested):
Please clean up the PR, there are old, already merged, commits in there.
Can you come up with a better commit title? Don't forget prefixing with `arcv:` and using `--signoff`.

PR #192, opened 2025-11-28
ARC-V RHX-100 upstream patch series

MichielDerhaeg:


gcc/config/riscv/riscv.cc
```diff
@@ -10573,30 +10574,6 @@ riscv_store_data_bypass_p (rtx_insn *out_insn, rtx_insn *in_insn)
   return store_data_bypass_p (out_insn, in_insn);
 }
 
-/* Implement one boolean function for each of the values of the
```
MichielDerhaeg, line 11197: Should the declarations in the header be moved to arcv.h?

gcc/config/riscv/riscv-protos.h
```diff
@@ -860,6 +860,9 @@ extern rtx riscv_prefetch_cookie (rtx, rtx);
 extern bool riscv_prefetch_offset_address_p (rtx, machine_mode);
 
 struct riscv_tune_param;
+
+extern int riscv_get_tune_param_issue_rate (void);
```
MichielDerhaeg: This causes a circular dependency between riscv.cc and arcv.cc. Can we instead just pass in the issue_rate itself as a parameter where this is called?

gcc/config/riscv/riscv.cc
```diff
@@ -4392,7 +4392,8 @@ riscv_rtx_costs (rtx x, machine_mode mode, int outer_code, int opno ATTRIBUTE_UN
 	}
       gcc_fallthrough ();
     case SIGN_EXTRACT:
-      if (TARGET_XTHEADBB && outer_code == SET
+      if ((TARGET_ARCV_RHX100 || TARGET_XTHEADBB)
+	  && outer_code == SET
```
MichielDerhaeg, line 4716: FYI, this was added for the bit-extract fusion.

gcc/config/riscv/riscv.md
```diff
@@ -4536,6 +4595,68 @@
...
+     && (INTVAL (operands[2]) > 1 || !TARGET_ZBS)"
+  {
+     int amount = INTVAL (operands[2]);
+     int end = INTVAL (operands[3]) + amount;
+     operands[2] = GEN_INT (BITS_PER_WORD - end);
+     operands[3] = GEN_INT (BITS_PER_WORD - amount);
+     return "slli\t%0,%1,%2\n\tsrli\t%0,%0,%3";
+  }
+  [(set_attr "type" "alu_fused")]
+)
```
MichielDerhaeg: As far as I can tell, this fusion was never implemented as a define_insn_and_split. Might not be trivial to force these exact instructions after a split.

MichielDerhaeg: All of these patterns used to be a defined_insn_and_split. In case we can't get rid of them, let's try to go back to define_insn_and_split (check the downstream commit history) and see what the performance looks like.

MichielDerhaeg (commented):
I tried to split up the commits in something sensible. Didn't check whether they can be built individually though.

gcc/config/riscv/riscv.cc
luismgsilva: I know I originally placed the declarations in `arcv.h`, but looking at the existing code structure, it may be a better fit to keep them in `riscv-protos.h`

EDIT: See https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/192/commits/d479345f57d82fd23d84f5778ad6b94ad23e439b

gcc/config/riscv/riscv-protos.h
luismgsilva: Indeed. Yes.

EDIT: See https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/192/commits/73219a427a993c2657bd3189054da79516dbd4b4

gcc/config/riscv/riscv.md
luismgsilva: If we revert `madd_split_fused1` back to a `define_insn_and_split` we get 0.898% improvement.

MichielDerhaeg: I ran embench before and after the revert to `define_insn_and_split`.
There were some regressions and improvements but the numbers were mostly stable except for the matmul benchmark.
Biggest regression was edn with a 2% drop.
Largest improvement was matmult with a 12% improvements. So I think we can keep this change.

MichielDerhaeg: `madd_split_fused_extended` also still needs to be converted to a `define_insn_and_split`.

luismgsilva: > `madd_split_fused_extended` also still needs to be converted to a `define_insn_and_split`.

Reverted in: https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/192/commits/191282c0e6a85b91411b0863cb280134deeb7a9f

gcc/config/riscv/riscv.md
```diff
@@ -4633,17 +4633,23 @@
...
        {
-	 return "mulw\t%0,%1,%2\n\taddw\t%0,%0,%3";
+        emit_insn (gen_mulsi3_extended (operands[0], operands[1], operands[2]));
+        emit_insn (gen_addsi3_extended (operands[0], operands[0], operands[3]));
        }
-  }
-  [(set_attr "type" "imul_fused")]
+    DONE;
+   }"
+  [(set_attr "type" "imul")]
```
MichielDerhaeg: We might want to keep imul_fused.

gcc/config/riscv/riscv.md
```diff
@@ -4633,17 +4633,23 @@
...
-  {
+  "#"
+  "&& reload_completed"
+  [(const_int 0)]
+  "{
      if (REGNO (operands[0]) == REGNO (operands[3]))
        {
-	 return "mulw\t%4,%1,%2\n\taddw\t%4,%3,%4\n\tmv\t%0,%4";
+        emit_insn (gen_mulsi3_extended (operands[4], operands[1], operands[2]));
+        emit_insn (gen_addsi3_extended (operands[0], operands[3], operands[4]));
```
MichielDerhaeg: Both target operand registers need to be the same or the fusion won't happen.
We do need to emit the move here.

gcc/config/riscv/riscv.md
```diff
@@ -4536,6 +4595,81 @@
...
+        emit_insn (gen_addsi3 (operands[0], operands[3], operands[4]));
+       }
+     else
+       {
+        emit_insn (gen_mulsi3 (operands[0], operands[1], operands[2]));
+        emit_insn (gen_addsi3 (operands[0], operands[0], operands[3]));
+       }
+    DONE;
+   }"
+  [(set_attr "type" "imul")]
```
luismgsilva: Should this also be as follows?
```suggestion
  [(set_attr "type" "imul_fused")]
```

gcc/config/riscv/riscv.md
```diff
@@ -4610,13 +4610,13 @@
   "{
      if (REGNO (operands[0]) == REGNO (operands[3]))
        {
-        emit_insn (gen_mulsi3 (operands[4], operands[1], operands[2]));
-        emit_insn (gen_addsi3 (operands[0], operands[3], operands[4]));
+	emit_insn (gen_mulsi3 (operands[4], operands[1], operands[2]));
+	emit_insn (gen_addsi3 (operands[0], operands[3], operands[4]));
```
MichielDerhaeg: Ah, this is also wrong.

luismgsilva: Thanks. Fixed.

luismgsilva: Interesting with this fix https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/commit/b4ce3f98a95dfb5dc26e5c5bd0aa7449ca7e0553 I get a regression of 0.5%

gcc/config/riscv/riscv-protos.h
```diff
@@ -875,6 +875,7 @@ extern rtx riscv_prefetch_cookie (rtx, rtx);
 extern bool riscv_prefetch_offset_address_p (rtx, machine_mode);
 
 struct riscv_tune_param;
+
```
MichielDerhaeg: Remove

gcc/config/riscv/arcv.cc
```diff
@@ -522,6 +548,221 @@ arcv_macro_fusion_pair_p (rtx_insn *prev, rtx_insn *curr)
...
+
+  return sched_state.cached_can_issue_more;
+}
+
+int
+arcv_sched_adjust_priority (rtx_insn *insn, int priority)
+{
+  if (!TARGET_ARCV_RHX100)
+    return priority;
+
```
MichielDerhaeg: We already check whether the fusion was enabled in the caller.

luismgsilva: Do we want to check `RISCV_FUSE_ARCV` or specifically RHX100 ? I know currently it's the same think, but for future purposes.

gcc/config/riscv/riscv-protos.h
```diff
@@ -839,9 +839,18 @@ extern const char *th_output_move (rtx, rtx);
...
 extern bool arcv_mpy_2c_bypass_p (rtx_insn *, rtx_insn *);
 extern bool arcv_mpy_10c_bypass_p (rtx_insn *, rtx_insn *);
+extern bool arcv_macro_fusion_pair_p (rtx_insn *, rtx_insn *);
+extern void arcv_sched_fusion_priority (rtx_insn *, int, int *, int *);
+extern bool arcv_can_issue_more_p (int, int);
+extern int arcv_sched_variable_issue (rtx_insn *, int);
+extern void arcv_sched_init (void);
+extern int arcv_sched_reorder2 (rtx_insn **, int *);
+extern int arcv_sched_adjust_priority (rtx_insn *, int);
+extern int arcv_sched_adjust_cost (rtx_insn *, int, int);
```
MichielDerhaeg: These need to be added in the scheduling pass.

gcc/config/riscv/arcv.cc
MichielDerhaeg: I think we want most things to be gated by `RISCV_FUSE_ARCV` except for the things added by the scheduling patch (that adds reorder2, etc).
The scheduling patch adds things that are really specific about the pipeline definition but everything else is just about the fusion schedule grouping.

gcc/config/riscv/riscv-protos.h
luismgsilva: I might have messed up during rebasing. Fixing it...

luismgsilva: Fixed. Thanks.

luismgsilva:
@MichielDerhaeg Take a look at the current fixups

gcc/config/riscv/riscv-fusion.cc
```diff
@@ -758,5 +1001,209 @@ riscv_macro_fusion_pair_p (rtx_insn *prev, rtx_insn *curr)
...
+      /* Avoid duplicating ZEXTW (32/32) and ZEXTH (48/48) patterns
+	 which are already handled by upstream code.  */
+      if (!((shift_left == 32 && shift_right == 32)
+	    || (shift_left == 48 && shift_right == 48)))
+       {
+	 if (dump_file)
+	   fprintf (dump_file, "RISCV_FUSE_SHIFT_BITEXTRACT\n");
+	 return true;
+       }
+    }
```
MichielDerhaeg: As far as I can tell, this is just identical to `RISCV_FUSE_BFEXT`

gcc/config/riscv/riscv-fusion.cc
```diff
@@ -126,6 +162,213 @@ riscv_set_is_shNadduw (rtx set)
...
+  if (get_attr_type (prev) == TYPE_LOAD)
+    return REGNO (SET_DEST (prev_set)) != REGNO (SET_DEST (curr_set));
+
+  /* Check for store + LUI fusion (always allowed).  */
+  if (get_attr_type (prev) == TYPE_STORE)
+    return true;
+
+  return false;
+}
+
```
MichielDerhaeg, line 464: Upstream covers ld+lui already.
We can split this up so we just have to add st+lui.

gcc/config/riscv/riscv-fusion.cc
```diff
@@ -758,5 +1001,209 @@ riscv_macro_fusion_pair_p (rtx_insn *prev, rtx_insn *curr)
...
+	 return true;
+       }
+      if (riscv_memop_lui_pair_p (curr, prev))
+       {
+	 if (dump_file)
+	   fprintf (dump_file, "RISCV_FUSE_MEMOP_LUI (curr, prev)\n");
+	 return true;
+       }
+    }
+
```
MichielDerhaeg: If we merge part of this fusion with the upstream version. We can outline the upstream version and add a `_REV` fusion type for the reversed case.

gcc/config/riscv/riscv-fusion.cc
```diff
@@ -758,5 +1001,209 @@ riscv_macro_fusion_pair_p (rtx_insn *prev, rtx_insn *curr)
...
+	   fprintf (dump_file, "RISCV_FUSE_LS_UPDATE (prev, curr)\n");
+	 return true;
+       }
+      if (riscv_ls_update (curr, prev))
+       {
+	 if (dump_file)
+	   fprintf (dump_file, "RISCV_FUSE_LS_UPDATE (curr, prev)\n");
+	 return true;
+       }
+    }
```
MichielDerhaeg: There are upstream ones that looks similar. Have you checked what the exact differences are? (except for the reversibility)

gcc/config/riscv/riscv-fusion.cc
luismgsilva: Youre correct. Please see: https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/192/changes/6084a0edbae7e1b8bad69a1b92db25fb5c4c71c4

gcc/config/riscv/riscv-fusion.cc
luismgsilva: Ive simplified and removed the function call. I believe its easier to read like this: https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/192/changes/5d4f626891fed8f49e39f1d645474ca1f238d178

gcc/config/riscv/riscv-fusion.cc
luismgsilva: I have created a new `RISCV_FUSE_LUI_LD_REV` alongside the other `RISCV_FUSE_LUI_LD` See https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/192/changes/5ffb20e41305247cd9c35a8656d693c421f47b4e

But, shouldnt we align with the others? Either adapt this to use `(curr, prev)` or make it more modular and create new fusion types for the reversed cases. e.g., RISCV_FUSE_STORE_LUI and RISCV_FUSE_STORE_LUI_REV

gcc/config/riscv/riscv-fusion.cc
luismgsilva: Which ones are you referring to ?

MichielDerhaeg:
@exur00 not sure if this happened already, but the next_fusible_insn has to be fixed here as well

exur00:
> @exur00 not sure if this happened already, but the next_fusible_insn has to be fixed here as well

I have added it

gcc/config/riscv/riscv-fusion.cc
MichielDerhaeg: e.g. `RISCV_FUSE_ALIGNED_STD`

luismgsilva: ALIGNED_STD is stores-only. We could only add the load-only version.

But alignment requirement is different. ALIGNED_STD requires a lower offset to be aligned to `2 * mode_size`, whereas we only need each individual access to be aligned (`offset % mode_size`), so ALIGNED_STD would reject valid pairs. e.g., two sw at offsets 4 and 8 off the same base.

RISCV_FUSE_ALIGNED_STD:
```
	      /* We've normalized, so we need to check that the lower
		 address is aligned to 2X the size of the object.  The
		 higher address must be the lower address plus the
		 size of the object.  */
	      if (((offset_prev_int % (2 * mode_size)) == 0)
		  && offset_prev_int + mode_size == offset_curr_int)
```

RISCV_FUSE_ADJACENT_STORE:
```
  /* Fuse adjacent aligned addresses.  */
  if ((off0 % GET_MODE_SIZE (GET_MODE (mem0)).to_constant () == 0)
      && (abs (off1 - off0) == GET_MODE_SIZE (GET_MODE (mem0)).to_constant ()))
    return true;
```

Another thing is that ALIGNED_STD is gated on SCALAR_INT_MODE_P, which for now is fine. But we do have non-scalar modes in the future as per your PR https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/248.

```
	  && SCALAR_INT_MODE_P (GET_MODE (SET_DEST (curr_set)))
```

gcc/config/riscv/riscv-fusion.cc
MichielDerhaeg: I would outline the original fusions, and just reverse the input params for _rev.

gcc/config/riscv/riscv-fusion.cc
MichielDerhaeg: Okay, let's keep ours.

gcc/config/riscv/riscv-protos.h
```diff
@@ -837,7 +837,7 @@ enum riscv_fusion_pairs
   RISCV_FUSE_ADJACENT_LOAD = (1 << 16),
   RISCV_FUSE_ADJACENT_STORE = (1 << 17),
   RISCV_FUSE_LS_UPDATE = (1 << 18),
-  RISCV_FUSE_MEMOP_LUI = (1 << 19),
+  RISCV_FUSE_STORE_LUI = (1 << 19),
```
MichielDerhaeg: I would call it `RISCV_FUSE_LUI_ST` to align with the old ones.

gcc/testsuite/gcc.target/riscv/arcv-fusion-store-lui-dump.c
```diff
@@ -0,0 +1,20 @@
...
+  *p = val;
+  g1 = val;
+  p[1] = val;
+  g2 = val;
+  p[2] = val;
+  g3 = val;
+}
+
+/* { dg-final { scan-rtl-dump "RISCV_FUSE_STORE_LUI \\(prev, curr\\)" "sched2" } } */
+/* { dg-final { scan-rtl-dump "RISCV_FUSE_STORE_LUI \\(curr, prev\\)" "sched2" } } */
```
MichielDerhaeg: Why the changes here?

gcc/config/riscv/riscv-fusion.cc
```diff
@@ -0,0 +1,1204 @@
...
+      if (REG_P (store_src) && REG_P (load_dest)
+	  && REGNO (store_src) == REGNO (load_dest))
+       {
+	if (dump_file)
+	  {
+	    if (GET_MODE (store_src) != GET_MODE (load_dest))
+	      fprintf (dump_file, "RISCV_FUSE_LI_STORE (subreg)\n");
+	    else
+	      fprintf (dump_file, "RISCV_FUSE_LI_STORE\n");
+	  }
```
MichielDerhaeg: What's the point of this?

gcc/config/riscv/riscv-fusion.cc
```diff
@@ -0,0 +1,1204 @@
...
+      && get_attr_type (prev) == TYPE_MOVE
+      && get_attr_move_type (prev) == MOVE_TYPE_CONST
+      && get_attr_type (curr) == TYPE_STORE)
+    {
+      rtx store_src = SET_SRC (curr_set);
+      rtx load_dest = SET_DEST (prev_set);
+
+      if (REG_P (store_src) && REG_P (load_dest)
+	  && REGNO (store_src) == REGNO (load_dest))
+       {
```
MichielDerhaeg: I would merge all of the conditions in a single if statement like the other ones.

it is fine to repeat `SET_SRC (curr_set)` and such.

gcc/testsuite/gcc.target/riscv/arcv-fusion-store-lui-dump.c
luismgsilva: `RISCV_FUSE_MEMOP_LUI` used to handle both load and store LUI fusion. When we split them apart, the existing test case didnt cover `RISCV_FUSE_LUI_ST` (previously `RISCV_FUSE_STORE_LUI`), so I added a test case for it.

gcc/config/riscv/riscv-fusion.cc
luismgsilva: This was already in the original code. Just distinguishes in the dump whether the `li` and `store` operate on the same mode or not. (e..g, `li` sets a `DImode` reg but the store uses it as `SImode`). Happy to drop the distinction and just print `RISCV_FUSE_LI_STORE` if you think it's just noise.

MichielDerhaeg: It can go, I think it isn't useful.

gcc/config/riscv/riscv-fusion.cc
```diff
@@ -741,6 +1222,22 @@ static const struct riscv_fusion_entry riscv_fusion_table[] =
...
+    riscv_fuse_li_branch, "RISCV_FUSE_LI_BRANCH" },
+  { RISCV_FUSE_ADJACENT_LOAD,
+    riscv_fuse_adjacent_load, "RISCV_FUSE_ADJACENT_LOAD" },
+  { RISCV_FUSE_ADJACENT_STORE,
+    riscv_fuse_adjacent_store, "RISCV_FUSE_ADJACENT_STORE" },
+  { RISCV_FUSE_LS_UPDATE,
+    riscv_fuse_ls_update, "RISCV_FUSE_LS_UPDATE" },
+  { RISCV_FUSE_LUI_ST,
+    riscv_fuse_lui_st, "RISCV_FUSE_LUI_ST" },
+  { RISCV_FUSE_LI_STORE,
```
luismgsilva, line 1287: Should we keep `_STORE` or `_ST` ?

gcc/config/riscv/arcv.cc
```diff
@@ -0,0 +1,417 @@
...
+  {
+    rtx_insn *insn = ready[*n_readyp - 1];
+    enum attr_type insn_type = get_attr_type (insn);
+
+    /* Memory operations go to pipeB if available.  */
+    if (!sched_state.pipeB_scheduled_p
+       && (insn_type == TYPE_LOAD || insn_type == TYPE_STORE))
+    {
+      sched_state.pipeB_scheduled_p = 1;
+      sched_state.cached_can_issue_more = 1;
```
luismgsilva: For some reason `sched_state.cached_can_issue_more = 1` got lost even in our old branch. I added it back in.

gcc/config/riscv/riscv-fusion.cc
```diff
@@ -691,6 +954,270 @@ riscv_fuse_b_alui (rtx_insn *prev, rtx_insn *curr)
...
+      && REG_P (SET_DEST (prev_set))
+      && REG_P (SET_DEST (curr_set))
+      && (GET_CODE (SET_SRC (curr_set)) == HIGH
+	  || (CONST_INT_P (SET_SRC (curr_set))
+	      && LUI_NONZERO_OPERAND (INTVAL (SET_SRC (curr_set)))))
+      && REGNO (SET_DEST (prev_set)) != REGNO (SET_DEST (curr_set)))
+    return true;
+
+  return false;
+}
```
MichielDerhaeg, line 1221: Sorry, I was the person that asked for this to be split. But our lui_ld fusion requires the destination registers to be different, but the one that is already upstream requires them to be the same, so we should merge them again I think and give it some other name.

gcc/config/riscv/riscv-fusion.cc
```diff
@@ -741,6 +1268,22 @@ static const struct riscv_fusion_entry riscv_fusion_table[] =
     riscv_fuse_bfext, "RISCV_FUSE_BFEXT" },
```
MichielDerhaeg: Also, BFEXT is not entirely correct, it matches too much. Jin Ma's patch checks something similar.

commit-msg/0001-RISC-V-Implement-riscv_macro_fusion_pair_p-for-ARC-V
```diff
@@ -0,0 +1,96 @@
...
+	* gcc.target/riscv/arcv-fusion-lui-ld-dump.c: New test.
+	* gcc.target/riscv/arcv-fusion-mult-add-op0-dump.c: New test.
+	* gcc.target/riscv/arcv-fusion-mult-add-op1-dump.c: New test.
+	* gcc.target/riscv/arcv-fusion-shift-bitextract-dump.c: New test.
+	* gcc.target/riscv/arcv-fusion-store-lui-dump.c: New test.
+
+Co-authored-by: Artemiy Volkov <artemiyv@acm.org>
+Co-authored-by: Michiel Derhaeg <michiel.derhaeg@globalfoundries.com>
+Signed-off-by: Luis Silva <luis.silva@globalfoundries.com>
+
```
MichielDerhaeg: Loeka needs to creditted

commit-msg/0001-RISC-V-Implement-riscv_macro_fusion_pair_p-for-ARC-V
```diff
@@ -0,0 +1,96 @@
...
+	* gcc.target/riscv/arcv-fusion-li-branch-dump.c: New test.
+	* gcc.target/riscv/arcv-fusion-li-store-dump.c: New test.
+	* gcc.target/riscv/arcv-fusion-ls-update-dump.c: New test.
+	* gcc.target/riscv/arcv-fusion-ls-update-move-dump.c: New test.
+	* gcc.target/riscv/arcv-fusion-ls-update-rev-dump.c: New test.
+	* gcc.target/riscv/arcv-fusion-lui-ld-dump.c: New test.
+	* gcc.target/riscv/arcv-fusion-mult-add-op0-dump.c: New test.
+	* gcc.target/riscv/arcv-fusion-mult-add-op1-dump.c: New test.
+	* gcc.target/riscv/arcv-fusion-shift-bitextract-dump.c: New test.
+	* gcc.target/riscv/arcv-fusion-store-lui-dump.c: New test.
```
MichielDerhaeg: I put it on the confluence page, and it is a bit of nit. But I would simplify the test names.
just call it `arcv-fusion-<fusion_name>-<number>.c` I don't know where the `dump` comes from, that's not really gcc convention I think.

Ah, also, not call it `arcv-`? what are the other fusion tests called?

commit-msg/0001-RISC-V-Implement-riscv_macro_fusion_pair_p-for-ARC-V
```diff
@@ -0,0 +1,96 @@
+From 13824e9e44a82d37a9b67de0c15ddec3d52eb5b3 Mon Sep 17 00:00:00 2001
+From: Luis Silva <luis.silva@globalfoundries.com>
+Date: Mon, 13 Jul 2026 13:16:01 +0100
+Subject: [PATCH 1/4] RISC-V: Implement riscv_macro_fusion_pair_p for ARC-V
+ RHX-100 series.
+
+This patch adds the macro-op fusion cases used by the ARC-V RHX-100 and
+enables them in the tune_info (fusible_ops).  Fusion rules are collected in a
+table and dispatched from riscv_macro_fusion_pair_p; each is gated by its own
+RISCV_FUSE_* flag so a tune enables only the pairs its hardware supports.
```
MichielDerhaeg: > Fusion rules are collected in a
> table and dispatched from riscv_macro_fusion_pair_p; each is gated by its own
> RISCV_FUSE_* flag so a tune enables only the pairs its hardware supports.

Is obvious IMO, can be removed.

commit-msg/0001-RISC-V-Implement-riscv_macro_fusion_pair_p-for-ARC-V
```diff
@@ -0,0 +1,96 @@
...
+
+Adjacent load/store fusion is prioritized: the other pairs are suppressed
+until sched2 so double load/store opportunities are not broken up.
+
+Fusion cases enabled for the RHX-100 (I1 -> I2):
```
MichielDerhaeg: You still reference I1/2 from the spec, but not in any of the descriptions.

commit-msg/0001-RISC-V-Implement-riscv_macro_fusion_pair_p-for-ARC-V
```diff
@@ -0,0 +1,96 @@
...
+  RISCV_FUSE_LUI_LD_REV     (riscv_fuse_lui_ld_rev)
+      load + lui (reversed lui+ld), with distinct destinations.
+  RISCV_FUSE_LS_UPDATE      (riscv_fuse_ls_update)
+      load + addi, or addi + store, updating the base register.
+  RISCV_FUSE_BFEXT          (riscv_fuse_bfext)
+      slli rd,rs + srli rd,rd (bitfield extract).
+  RISCV_FUSE_ADJACENT_LOAD  (riscv_fuse_adjacent_load)
+      two loads, same base, offsets |off1-off2| == XLEN/8.
+  RISCV_FUSE_ADJACENT_STORE (riscv_fuse_adjacent_store)
+      two stores, same base, offsets |off1-off2| == XLEN/8.
```
MichielDerhaeg: We need to indicate which ones are reversible.

commit-msg/0001-RISC-V-Implement-riscv_macro_fusion_pair_p-for-ARC-V
```diff
@@ -0,0 +1,96 @@
...
+
+riscv_fuse_bfext and riscv_fuse_lui_ld already existed; this patch adds the
+other eight checkers and enables the full set above for the RHX-100.  The
+memory pairs only fire after reload (reload_completed && !sched_fusion), which
+is why they are deferred to sched2.
```
MichielDerhaeg: Unfortunately not true

commit-msg/0001-RISC-V-Implement-riscv_macro_fusion_pair_p-for-ARC-V
```diff
@@ -0,0 +1,96 @@
...
+    separate arcv.cc with a monolithic arcv_macro_fusion_pair_p; each
+    pair is now an individual riscv_fuse_* checker.
+  - Replaced the single RISCV_FUSE_ARCV umbrella flag with per-pair
+    RISCV_FUSE_* flags, enabled via the RHX-100 tune_info fusible_ops.
+  - Reuse the existing RISCV_FUSE_BFEXT and RISCV_FUSE_LUI_LD checkers
+    instead of the custom shift-bitextract / load+lui logic from v1.
+  - Restructured the v1 memop+arith / memop+lui helpers into the more
+    specific LS_UPDATE, LUI_ST and LUI_LD_REV checkers.
+  - Dropped the config.gcc / t-riscv / arcv.o plumbing (no new file).
+  - Added 12 dump-scan tests (none in v1).
```
MichielDerhaeg: Also mention that many checks were added (like `REG_P`) that were forgotten before.

commit-msg/0003-RISC-V-Implement-scheduling-for-ARC-V-RHX-100-series
```diff
@@ -0,0 +1,85 @@
...
+
+	* gcc.target/riscv/arcv-fusion-ls-update-dump.c: Remove XFAIL now
+	that RHX-100 scheduling is implemented.
+	* gcc.target/riscv/arcv-fusion-ls-update-rev-dump.c: Likewise, and
+	fix the scan-rtl-dump pattern.
+
+Co-authored-by: Artemiy Volkov <artemiyv@acm.org>
+Co-authored-by: Michiel Derhaeg <michiel.derhaeg@globalfoundries.com>
+Co-authored-by: Alex Turjan <alex.turjan@globalfoundries.com>
+Signed-off-by: Luis Silva <luis.silva@globalfoundries.com>
```
MichielDerhaeg, line 65: Loeka needs credits here also.

commit-msg/0003-RISC-V-Implement-scheduling-for-ARC-V-RHX-100-series
```diff
@@ -0,0 +1,85 @@
...
+
+===
+
+Changes since v1:
+  - Gate the scheduler hooks on TARGET_ARCV_RHX100 directly instead of
+    riscv_fusion_enabled_p (RISCV_FUSE_ARCV); dropped the RISCV_FUSE_ARCV
+    umbrella and the riscv_fusion_enabled_p forward declaration in
+    riscv.cc.
+  - Removed the custom arcv_next_fusible_insn helper and use the generic
+    next_nonnote_nondebug_insn_bb instead.
```
MichielDerhaeg: to exactly match the scheduler.

commit-msg/0003-RISC-V-Implement-scheduling-for-ARC-V-RHX-100-series
```diff
@@ -0,0 +1,85 @@
...
+  - Call the generic riscv_macro_fusion_pair_p (the table dispatcher)
+    instead of arcv_macro_fusion_pair_p, which no longer exists after the
+    fusion rework.
+  - arcv_sched_adjust_priority and arcv_sched_variable_issue now use
+    next_nonnote_nondebug_insn_bb + single_set instead of next_insn +
+    INSN_P, dropping the manual DEBUG/USE/CLOBBER guards.
```
MichielDerhaeg: you can merge that with the other comment about this, just mention we switched to this function and it is more correct. as it matches what the scheduler does.

commit-msg/0003-RISC-V-Implement-scheduling-for-ARC-V-RHX-100-series
```diff
@@ -0,0 +1,85 @@
...
+  - reorder2's single-instruction fallback now updates
+    cached_can_issue_more in both the memory and ALU branches.
+  - Dropped the bogus riscv.h ChangeLog entry: TARGET_ARCV_RHX100 is now
+    defined by the fusion patch, so this patch no longer touches riscv.h.
+  - Un-XFAIL'd arcv-fusion-ls-update-dump.c and
+    arcv-fusion-ls-update-rev-dump.c now that RHX-100 scheduling is
+    implemented (no test changes in v1).
```
MichielDerhaeg: Similar to the other patch, not enough checking was done previously and it should now be more correct.

commit-msg/0004-RISC-V-Add-instruction-patterns-for-32-bit-multiply-
```diff
@@ -0,0 +1,54 @@
+From ab7bc51b4d040fb24e3444d5d0faf705d873b0be Mon Sep 17 00:00:00 2001
+From: Luis Silva <luis.silva@globalfoundries.com>
+Date: Mon, 13 Jul 2026 13:46:34 +0100
+Subject: [PATCH 4/4] RISC-V: Add instruction patterns for 32-bit multiply-add
+ and bit-extract fusion.
```
MichielDerhaeg: even the title needs to changed here

gcc/config/riscv/arcv.cc
```diff
@@ -404,13 +404,6 @@ arcv_sched_variable_issue (rtx_insn *insn, int more)
 	sched_state.alu_pipe_scheduled_p = 1;
     }
 
-  if (get_attr_type (insn) == TYPE_ALU_FUSED
-      || get_attr_type (insn) == TYPE_IMUL_FUSED)
-    {
-      sched_state.alu_pipe_scheduled_p = 1;
-      more -= 1;
-    }
-
```
MichielDerhaeg: Oops, added to the wrong commit.

gcc/config/riscv/iterators.md
```diff
@@ -225,6 +225,8 @@
 				      (zero_extract "srliw")])
 (define_code_attr extract_shift [(sign_extract "ashiftrt")
 				 (zero_extract "lshiftrt")])
+(define_code_attr is_zero_extract [(sign_extract "false")
+				   (zero_extract "true")])
```
MichielDerhaeg: Forgot to remove this.

gcc/config/riscv/riscv-fusion.cc
luismgsilva: I do get a decrease in performance from `6.972939` to `6.941197` merging as it was.

EDIT: See https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/commit/3131bd33143885449ae4f84402e8bb91fd4cf61a

commit-msg/0001-RISC-V-Implement-riscv_macro_fusion_pair_p-for-ARC-V
luismgsilva: See https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/296

gcc/config/riscv/arcv.cc
MichielDerhaeg: I removed this in the original set of upstream patches because when you have 2 consequetive XBFU fusions:

```
slli <- cycle 1
srli <- cycle 1
slli <- cycle 1
srli <- cycle 2
```

The scheduler will start a new cycle after the second `slli` even though the hardware can run all 4 instructions in a single cycle.
This is fixed when these lines are removed, but it causes regressions elsewhere.
This is not a problem downstream because the xbfu fusion is still represented as a single instruction.

 I suggest we keep it like this (current state) upstream and fix it later.

gcc/config/riscv/riscv.md
```diff
@@ -4831,14 +4831,22 @@
   "&& 1"
   [(const_int 0)]
   "{
+     rtx addend = operands[3];
```
MichielDerhaeg, line 4881: This change regresses performance somewhat but it is required for correctness.

The regression is not caused by this pattern, but other inaccuracies in the scheduling.
I suggest to postpone dealing with that regression for now.

gcc/config/riscv/riscv.cc
```diff
@@ -4710,7 +4714,9 @@ riscv_rtx_costs (rtx x, machine_mode mode, int outer_code, int opno ATTRIBUTE_UN
 	}
       gcc_fallthrough ();
     case SIGN_EXTRACT:
-      if (TARGET_XTHEADBB && outer_code == SET
+      if ((TARGET_XTHEADBB
+	   || riscv_fusion_enabled_p (RISCV_FUSE_BFEXT))
```
luismgsilva: This change triggered these failures:
```
FAIL: gcc.target/riscv/ext-dce-promote-2.c   -O0   scan-assembler \\tsrliw\t
FAIL: gcc.target/riscv/ext-dce-promote-2.c   -O1   scan-assembler \\tsrliw\t
FAIL: gcc.target/riscv/ext-dce-promote-2.c   -O2   scan-assembler \\tsrliw\t
FAIL: gcc.target/riscv/ext-dce-promote-2.c   -O3 -g   scan-assembler \\tsrliw\t
FAIL: gcc.target/riscv/ext-dce-promote-2.c   -Os   scan-assembler \\tsrliw\t
FAIL: gcc.target/riscv/ext-dce-promote-2.c  -Og -g   scan-assembler \\tsrliw\t
FAIL: gcc.target/riscv/ext-dce-promote-2.c  -Oz   scan-assembler \\tsrliw\t
```

This change lowered the RTL cost of `zero_extract` \ `sign_extract` whenever `RISCV_FUSE_BFXT` is enabled. With RHX as default mtune, combine prefers extract-shaped RTL over the srliw paths in riscv.md. The test case requires `srliw`.

gcc/config/riscv/riscv.cc
```diff
@@ -16537,6 +16582,9 @@ riscv_memtag_tag_bitsize ()
 #undef TARGET_SCHED_MACRO_FUSION_PAIR_P
 #define TARGET_SCHED_MACRO_FUSION_PAIR_P riscv_macro_fusion_pair_p
 
+#undef TARGET_SCHED_FUSION_PRIORITY
+#define TARGET_SCHED_FUSION_PRIORITY riscv_sched_fusion_priority
```
luismgsilva, line 16665: We turned on GCC's sched_fusion pass for RISC-V by implementing `TARGET_SCHED_FUSION_PRIORITY`. The `interrupt-misaligned.c` test case checks that interrupt prologue/epilogue use a fixed order of stack saves (sd/sw/fsd at specific offsets). The test already disables normal instruction scheduling (-fno-schedule-insns, -fno-schedule-insns2) but not sched fusion, so the new pass reordered those memory operations.

I would add ` -fno-schedule-fusion` to the test case also.

gcc/config/riscv/riscv.cc
```diff
@@ -980,7 +980,11 @@ static const struct riscv_tune_param arcv_rhx100_tune_info = {
   false,				       /* overlap_op_by_pieces */
   true,					       /* use_zero_stride_load */
   false,				       /* speculative_sched_vsetvl */
-  RISCV_FUSE_NOTHING,			       /* fusible_ops */
+  (RISCV_FUSE_MULT_ADD | RISCV_FUSE_LI_BRANCH
+   | RISCV_FUSE_ADJACENT_LOAD | RISCV_FUSE_ADJACENT_STORE
+   | RISCV_FUSE_LS_UPDATE | RISCV_FUSE_LUI_ST
+   | RISCV_FUSE_LI_STORE | RISCV_FUSE_BFEXT
```
MichielDerhaeg: Regarding BFEXT. I suggest we simply split it into 2 separate fusions: one with srli and one with srai.

luismgsilva: Initial PR https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/305

gcc/config/riscv/riscv-protos.h
```diff
@@ -867,11 +867,21 @@ enum riscv_fusion_pairs
   RISCV_FUSE_BFEXT = (1 << 11),
   RISCV_FUSE_EXPANDED_LD = (1 << 12),
   RISCV_FUSE_B_ALUI = (1 << 13),
+  RISCV_FUSE_MULT_ADD = (1 << 14),
+  RISCV_FUSE_LI_BRANCH = (1 << 15),
+  RISCV_FUSE_ADJACENT_LOAD = (1 << 16),
+  RISCV_FUSE_ADJACENT_STORE = (1 << 17),
+  RISCV_FUSE_LS_UPDATE = (1 << 18),
+  RISCV_FUSE_LUI_ST = (1 << 19),
+  RISCV_FUSE_LI_STORE = (1 << 20),
```
luismgsilva, line 876: @MichielDerhaeg Should we rename `RISCV_FUSE_LI_STORE` to `RISCV_FUSE_LI_ST` to align with others ?

luismgsilva:
V2 Patches have been sent upstream: https://gcc.gnu.org/pipermail/gcc-patches/2026-August/726434.html

PR #193, opened 2025-12-12
Rmx500 latencies

Alexehv77:


gcc/config/riscv/riscv.cc
```diff
@@ -605,17 +605,17 @@ static const struct riscv_tune_param arcv_rmx100_tune_info = {
...
-/* Costs to use when optimizing for Synopsys RMX-500.  */
-static const struct riscv_tune_param arcv_rmx500_tune_info = {
-  {COSTS_N_INSNS (2), COSTS_N_INSNS (2)},	/* fp_add */
-  {COSTS_N_INSNS (2), COSTS_N_INSNS (2)},	/* fp_mul */
-  {COSTS_N_INSNS (17), COSTS_N_INSNS (17)},	/* fp_div */
-  {COSTS_N_INSNS (2), COSTS_N_INSNS (2)},	/* int_mul */
-  {COSTS_N_INSNS (17), COSTS_N_INSNS (17)},	/* int_div */
-  1,						/* issue_rate */
-  4,						/* branch_cost */
+/* Costs to use when optimizing for Synopsys RHX-100.  */
```
luismgsilva: Why did you change the `riscv_tune_param` order of `rmx500 -> rhx100 -> rpx100` to `rhx100 -> rpx100 -> rmx500` ?

gcc/config/riscv/riscv.cc
```diff
@@ -649,13 +649,13 @@ static const struct riscv_tune_param arcv_rhx100_tune_info = {
...
-  {COSTS_N_INSNS (4), COSTS_N_INSNS (5)},	/* fp_mul */
-  {COSTS_N_INSNS (20), COSTS_N_INSNS (20)},	/* fp_div */
+/* Costs to use when optimizing for Synopsys RMX-500.  */
+static const struct riscv_tune_param arcv_rmx500_tune_info = {
+  {COSTS_N_INSNS (4), COSTS_N_INSNS (64)},	/* fp_add */
+  {COSTS_N_INSNS (4), COSTS_N_INSNS (75)},	/* fp_mul */
+  {COSTS_N_INSNS (20), COSTS_N_INSNS (120)},	/* fp_div */
   {COSTS_N_INSNS (4), COSTS_N_INSNS (4)},	/* int_mul */
-  {COSTS_N_INSNS (27), COSTS_N_INSNS (43)},	/* int_div */
+  {COSTS_N_INSNS (27), COSTS_N_INSNS (87)},	/* int_div */
```
MichielDerhaeg: What do the higher numbers mean?

MichielDerhaeg (approved):
Measurements look good.
Don't forget `--signoff` before merging.

gcc/config/riscv/riscv.cc
Alexehv77: this tune_info datastructure is used by the rtx_costs hook. this hook returns costs for certain instructions (cycles or code size depending on last argument of the function hook). gcc has various ways to generate code for a given source code operation. this hook is then used to decide what option to choose for generating code (native hw/internal gcc emulation/library).  In case of the rmx500 you have there for example the costs for single or double fp div. the one with cost 20 is the native hw for 32 bit fp, while the 120 refers to the 64 bit fp which as far as i understand is not natively supported so must rely on a library call.

gcc/config/riscv/riscv.cc
Alexehv77: There is no special intention and the order doesnt influence performance. I can change it.

MichielDerhaeg (approved):

luismgsilva (approved):

PR #198, opened 2026-01-12
RHX100 backport

luismgsilva:
This patch series reverts the downstream RHX100 implementation and applies the upstreamed RHX100 patches (see [0]).

[0] https://gcc.gnu.org/pipermail/gcc-patches/2026-January/705022.html


gcc/config/riscv/arcv.cc
```diff
@@ -340,6 +340,10 @@ arcv_memop_lui_pair_p (rtx_insn *prev, rtx_insn *curr)
 bool
 arcv_macro_fusion_pair_p (rtx_insn *prev, rtx_insn *curr)
 {
+  /* Never create sched groups with more than 2 members.  */
+  if (SCHED_GROUP_P (prev))
+    return false;
+
   rtx prev_set = single_set (prev);
```
MichielDerhaeg: We could cherry-pick https://github.com/gcc-mirror/gcc/commit/83d19b5d842dadc1720b57486d4675a238966ba4#diff-97684bd8d2f96ae10d7d0f70b17331d4062ef762e79fcf33763afce2859d5107R10265-R10441 if it applies cleanly

luismgsilva: Sure. Although we hit an ICE at the `dep_fusion` pass during coremark. Looking into it.

MichielDerhaeg: That reminds me, there is a fix upstream that fixes an ICE for targets that don't implement the macro_op hook in dep_fusion. arc will likely break if we don't add it.

MichielDerhaeg: https://gcc.gnu.org/pipermail/gcc-patches/2025-September/694586.html

luismgsilva: I will apply it because of ARC-Classic, although for the current issue I encountered that's not the issue.

luismgsilva: ICE fixed with https://github.com/gcc-mirror/gcc/commit/7999eb8672c7a5bb6923b1df75dbf29d60637713

gcc/config/riscv/t-riscv
```diff
@@ -162,7 +162,7 @@ riscv-zicfilp.o: $(srcdir)/config/riscv/riscv-zicfilp.cc \
 	$(COMPILER) -c $(ALL_COMPILERFLAGS) $(ALL_CPPFLAGS) $(INCLUDES) \
 		$(srcdir)/config/riscv/riscv-zicfilp.cc
 
-arcv.o: $(srcdir)/config/riscv/arcv.cc \
+riscv-arcv.o: $(srcdir)/config/riscv/arcv.cc \
```
MichielDerhaeg, line 165: @kolerov you should probably be aware of this. We didn't make this change upstream because `libgloss/riscv/arcv.c` doesn't exist upstream.

gcc/config/riscv/riscv.h
```diff
@@ -996,6 +996,12 @@ extern enum riscv_cc get_riscv_cc (const rtx use);
 #define TARGET_ARCV_RHX100 \
   (riscv_microarchitecture == arcv_rhx100)
 
+/* True if the target is an ARC-V core with fusion support.  */
+#define TARGET_ARCV_FUSION \
+  ((riscv_microarchitecture == arcv_rmx500) \
+   || (riscv_microarchitecture == arcv_rhx100) \
+   || (riscv_microarchitecture == arcv_rpx100))
```
MichielDerhaeg, line 999: Should we use `riscv_fusion_enabled_p (RISCV_FUSE_ARCV)` instead?

MichielDerhaeg:
I would merge 90cb37ac6b522578b739e5a85bb69f52be8a1291 8aa88fbdaa7774d58ddb0b0d04c20b2b74ee497b and a9c9a0fc502900c85a57634f760b06d938e195a0 into 1 commit.

gcc/config/riscv/t-riscv
MichielDerhaeg: ~a9c9a0fc502900c85a57634f760b06d938e195a0~ 544bbbd503835d1244ca342e0e4090fd0e172cbb for context

gcc/config/riscv/riscv.h
luismgsilva: Both `riscv_fusion_enabled_p` and `RISCV_FUSE_ARCV` would not be declared in that scope.
I followed the same structure as other do. e.g., https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/blob/544bbbd503835d1244ca342e0e4090fd0e172cbb/gcc/config/riscv/riscv.h#L991

gcc/config/riscv/riscv.h
```diff
@@ -996,6 +996,12 @@ extern enum riscv_cc get_riscv_cc (const rtx use);
 #define TARGET_ARCV_RHX100 \
   (riscv_microarchitecture == arcv_rhx100)
```
MichielDerhaeg: This is no longer used.

gcc/config/riscv/riscv.h
```diff
@@ -996,6 +996,12 @@ extern enum riscv_cc get_riscv_cc (const rtx use);
...
 
+/* True if the target is an ARC-V core with fusion support.  */
+#define TARGET_ARCV_FUSION \
+  ((riscv_microarchitecture == arcv_rmx500) \
+   || (riscv_microarchitecture == arcv_rhx100) \
+   || (riscv_microarchitecture == arcv_rpx100))
+
```
MichielDerhaeg, line 1000: Should this be defined in terms of `riscv_fusion_enabled_p (RISCV_FUSE_ARCV)`? Or should we use one of the two? We're now mixing both without pattern.

luismgsilva: I think we should rather use the `TARGET_ARCV_FUSION` whenever we want check if fusion is enabled because otherwise, we would need to change enum definitions to `riscv.h` and changing function scopes, which would be unlikely to be accepted upstream.

MichielDerhaeg (approved):

PR #199, opened 2026-01-19
arcv: Remove nonexistant vnorm intrinsics with eew=8

exur00:


exur00:
I should still run all the tests and add a description, where can I reference the isa from?

MichielDerhaeg (approved):
Ok, after you add a description.

MichielDerhaeg:
> where can I reference the isa from?

no need, you can just mention that the spec explicitly disallows support for eew8 and that the resulting code wouldn't run

PR #202, opened 2026-01-29
loop2_unroll: split loop exit during unrolling of uncountable loops

luismgsilva:
Consider the (lightly modified) core_list_reverse () function from Coremark:

struct list_node *
core_list_reverse (struct list_node *list)
{
  struct list_node *next = 0, *tmp;
  #pragma GCC unroll 4
  while (list)
    {
      tmp = list->next;
      list->next = next;
      next = list;
      list = tmp;
    }

  return next;
}

On AArch64, this compiles to the following:

core_list_reverse:
        cbz     x0, .L2
        ldr     x1, [x0]
        mov     x6, 0
        str     x6, [x0]
        mov     x3, x0
        cbz     x1, .L2
.L4:
        ldr     x2, [x1]
        str     x3, [x1]
        mov     x0, x1
        cbz     x2, .L2
        ldr     x4, [x2]
        str     x1, [x2]
        mov     x0, x2
        cbz     x4, .L2
	...
        mov     x0, x5
        mov     x3, x0
        ldr     x1, [x0]
        str     x6, [x0]
        cbnz    x1, .L4
.L2:
        ret

The 'next' variable lives in the x0 register, which is maintained by the
"mov x0, xR" instruction at every unrolled iteration.  However, this can
be improved by removing the instruction and splitting the loop exit into
multiple exits, each corresponding to the hard register in which the
'next' variable is at a particular iteration, so that the code looks
like:

core_list_reverse:
        cbz     x0, .L2
        mov     x1, 0
.L3:
        ldr     x2, [x0]
        str     x1, [x0]
        cbz     x2, .L2
        ldr     x3, [x2]
        str     x0, [x2]
        cbz     x3, .L13
	...
        ldr     x0, [x1]
        str     x3, [x1]
        cbnz    x0, .L3
        mov     x0, x1
.L2:
        ret
.L13:
        mov     x0, x2
        ret
.L14:
        mov     x0, x3
        ret

This patch implements this transformation by splitting variables defined
in the loop and live at the (single) exit BB of an uncountable loop,
replacing each of those with a unique temporary pseudo inside the loop
and assigning these pseudos back to the original variables in the newly
split exit BBs (one per unrolled iteration).  (This is the behavior that
the GIMPLE unroller would exhibit were it capable of handling
uncountable loops.)  Afterwards, the cprop pass is able to propagate the
(split) loop variables and carry the move instructions out of the loop.

This change is primarily intended for small in-order cores on which the
latency of the move isn't hidden by the latency of the load.  The
optimization is guarded by the new -fsplit-exit-in-unroller flag that is
off by default.  The flag has been documented in doc/invoke.texi, and the
common.opt.urls file has been regenerated.

The change has been bootstrapped and regtested on i386, x86_64, and
aarch64, and additionally regtested on riscv32.  Two new testcases have
been added to demonstrate operation and interaction with
-fvariable-expansion-in-unroller.  On a Cortex-A53, this patch leads to
an ~0.5% improvement for Coremark and SPECINT2006 geomean (the only
regression being 483.xalancbmk), when compiled with -O2
-funroll-all-loops.  The compile-time increase is ~0.4%.

---
Taken from: https://gcc.gnu.org/pipermail/gcc-patches/2025-April/681722.html

### DIFF
--- a/gcc/common.opt
+++ b/gcc/common.opt
@@ -2945,6 +2945,10 @@ fsingle-precision-constant
 Common Var(flag_single_precision_constant) Optimization
 Convert floating point constants to single precision constants.
 
+fsplit-exit-in-unroller
+Common Var(flag_split_exit_in_unroller) Init(1) Optimization
+Split exit basic blocks of uncountable loops.
+
 fsplit-ivs-in-unroller
 Common Var(flag_split_ivs_in_unroller) Init(1) Optimization
 Split lifetimes of induction variables when loops are unrolled.
--- a/gcc/common.opt
+++ b/gcc/common.opt
@@ -2946,7 +2946,7 @@ Common Var(flag_single_precision_constant) Optimization
 Convert floating point constants to single precision constants.
 
 fsplit-exit-in-unroller
-Common Var(flag_split_exit_in_unroller) Init(1) Optimization
+Common Var(flag_split_exit_in_unroller) Init(0) Optimization
 Split exit basic blocks of uncountable loops.
 
 fsplit-ivs-in-unroller
--- a/gcc/common.opt.urls
+++ b/gcc/common.opt.urls
@@ -1328,6 +1328,9 @@ UrlSuffix(gcc/Optimize-Options.html#index-fno-signed-zeros)
 fsingle-precision-constant
 UrlSuffix(gcc/Optimize-Options.html#index-fsingle-precision-constant)
 
+fsplit-exit-in-unroller
+UrlSuffix(gcc/Optimize-Options.html#index-fsplit-exit-in-unroller)
+
 fsplit-ivs-in-unroller
 UrlSuffix(gcc/Optimize-Options.html#index-fsplit-ivs-in-unroller)
 
--- a/gcc/doc/invoke.texi
+++ b/gcc/doc/invoke.texi
@@ -637,8 +637,8 @@ Objective-C and Objective-C++ Dialects}.
 -fsel-sched-pipelining  -fsel-sched-pipelining-outer-loops
 -fsemantic-interposition  -fshrink-wrap  -fshrink-wrap-separate
 -fsignaling-nans
--fsingle-precision-constant  -fsplit-ivs-in-unroller  -fsplit-loops
--fsplit-paths
+-fsingle-precision-constant -fsplit-exit-in-unroller  -fsplit-ivs-in-unroller
+-fsplit-loops -fsplit-paths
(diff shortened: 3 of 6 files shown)


MichielDerhaeg (approved):
Fine to merge if the new tests succeed

luismgsilva:
> Fine to merge if the new tests succeed

```
PASS: gcc.dg/loop-exit-split-1.c (test for excess errors)
PASS: gcc.dg/loop-exit-split-1.c scan-rtl-dump-times loop2_unroll "15:.*: optimized: loop unrolled 3 times" 1
PASS: gcc.dg/loop-exit-split-1.c scan-rtl-dump-times loop2_unroll ";; Found definition of \\(reg [0-9]+ \\[ list \\]\\)" 1
PASS: gcc.dg/loop-exit-split-1.c scan-rtl-dump-times loop2_unroll ";; Split exit edge " 4
PASS: gcc.dg/loop-exit-split-1.c scan-rtl-dump-times loop2_unroll ";; Splitting loop exit for loop 1" 1
PASS: gcc.dg/loop-exit-split-1.c scan-rtl-dump-times loop2_unroll ";; added [0-9]+ to list of pseudos to replace" 1
PASS: gcc.dg/loop-exit-split-1.c scan-rtl-dump-times loop2_unroll ";; replacing DU-chain for reg [0-9]+" 3
PASS: gcc.dg/loop-exit-split-2.c (test for excess errors)
PASS: gcc.dg/loop-exit-split-2.c scan-rtl-dump-times loop2_unroll "18:.*: optimized: loop unrolled 3 times" 1
PASS: gcc.dg/loop-exit-split-2.c scan-rtl-dump-times loop2_unroll ";; Found definition of \\(reg [0-9]+" 2
PASS: gcc.dg/loop-exit-split-2.c scan-rtl-dump-times loop2_unroll ";; Split exit edge " 4
PASS: gcc.dg/loop-exit-split-2.c scan-rtl-dump-times loop2_unroll ";; Splitting loop exit for loop 1" 1
PASS: gcc.dg/loop-exit-split-2.c scan-rtl-dump-times loop2_unroll ";; added [0-9]+ to list of pseudos to replace" 2
PASS: gcc.dg/loop-exit-split-2.c scan-rtl-dump-times loop2_unroll ";; replacing DU-chain for reg [0-9]+" 4
PASS: gcc.dg/loop-exit-split-2.c scan-rtl-dump-times loop2_unroll "Expanding Accumulator" 1

                === gcc Summary ===

# of expected passes            15
```

PR #205, opened 2026-02-04
arcv: Fix LMUL type mismatch in arcv_vmv_v_s intrinsics.

luismgsilva:
According to the latest spec, the __riscv_arcv_vmv_v_s_* intrinsics incorrectly required the vs2 operand to match the destination LMUL.  Per the instruction semantics, vs2 should always be LMUL=1 regardless of destination LMUL.
```
This caused type errors like:
  vint8m2_t result = __riscv_arcv_vmv_v_s_i8m2(vd, vs2, rs1, vl);
                                                    ^~~
  error: incompatible type for argument 2
  note: expected 'vint8m2_t' but argument is of type 'vint8m1_t'
```
This patch changed v_sx_args to use RVV_BASE_lmul1_vector for the second argument and added subreg conversion in the expander to reinterpret LMUL=1 operands as the full LMUL mode required by the MD pattern.

### DIFF
--- a/gcc/config/riscv/riscv-vector-builtins-bases.cc
+++ b/gcc/config/riscv/riscv-vector-builtins-bases.cc
@@ -2247,7 +2247,7 @@ public:
       rtx vs1 = expand_normal (CALL_EXPR_ARG (e.exp, arg_offset++));
 
       e.add_input_operand (mode, vd);
-      e.add_input_operand (mode, vs2);
+      e.add_input_operand (e.arg_mode (1), vs2);
       e.add_input_operand (Pmode, vs1);
     }
 
--- a/gcc/config/riscv/arcv-vector.md
+++ b/gcc/config/riscv/arcv-vector.md
@@ -138,18 +138,18 @@
   [(set_attr "type" "arith")])
 
 (define_insn "@pred_arcv_vmv_v_s_scalar<V_VLSI:mode><P:mode>"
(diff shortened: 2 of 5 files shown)


gcc/config/riscv/riscv-vector-builtins-bases.cc
```diff
@@ -2246,6 +2246,12 @@ class arcv_vmv : public function_base
       rtx vs2 = expand_normal (CALL_EXPR_ARG (e.exp, arg_offset++));
       rtx vs1 = expand_normal (CALL_EXPR_ARG (e.exp, arg_offset++));
 
+      machine_mode vs2_mode = e.arg_mode (1);
+      /* If vs2 is LMUL1 but we need full LMUL, create a subreg
+	 to reinterpret it.  */
+      if (vs2_mode != mode)
+	vs2 = simplify_gen_subreg (mode, vs2, vs2_mode, 0);
+
```
MichielDerhaeg: Might the instruction be incorrectly defined? I think it needs to specify the lmul1 mode for that parameter.

luismgsilva: I tried adjusting the pattern to use `lmul1`, but I couldn’t get it to work. Do you have any suggestions on how the pattern should look?

MichielDerhaeg (approved):

PR #206, opened 2026-02-05
arcv: Fix arcv.vclr intrinsic signature.

luismgsilva:
The arcv.vclr instruction takes (imm, vl) in the base variant and
(vector, imm, vl) for tail-undisturbed.  We add a custom shape and
expand method to handle both cases.


gcc/config/riscv/riscv-vector-builtins-bases.cc
```diff
@@ -2325,16 +2325,36 @@ class arcv_vsneg : public function_base
...
+    e.add_input_operand (arg_offset++);
+
+    /* Tail policy.  */
+    e.add_input_operand (Pmode, gen_int_mode (
+      e.pred == PRED_TYPE_tu ? TAIL_UNDISTURBED : TAIL_ANY, Pmode));
+
+    e.add_input_operand (Pmode, gen_int_mode (MASK_ANY, Pmode));
+    e.add_input_operand (Pmode, gen_int_mode (avl_type::NONVLMAX, Pmode));
+
+    return e.generate_insn (code_for_pred_arcv_vclr (e.vector_mode ()));
```
MichielDerhaeg: This looks like a reimplementation of `function_expander::use_exact_insn`.
What is the reason you can't just call that and overload the right predicates?

MichielDerhaeg (changes requested):

PR #207, opened 2026-02-11
arcv: Fix complex multiplication scalar operations to use vector forms.

luismgsilva:
arcv: Fix complex scalar multiplication intrinsics to use LMUL=1 scalar operands.

    Update ARC-V complex multiplication instructions to follow the XARCV 1.8
    specification.  The scalar operand (passed in a vector register) now uses
    scalar-in-vector (.vs) form with LMUL=1, independent of the destination vector
    LMUL, replacing the previous scalar (.vx) form.

    For widening operations, the scalar operand must be double-truncated with
    LMUL=1.

    The following instructions are updated:
     - arcv.vscmul.vs (changed from .vx to .vs)
     - arcv.vscjmul.vs (changed from .vx to .vs)
     - arcv.vwscmul.vs (changed from .vx to .vs, scalar is LMUL=1)
     - arcv.vwscjmul.vs (changed from .vx to .vs, scalar is LMUL=1)
     - arcv.vwscmac.vs (changed from .vx to .vs, scalar is LMUL=1)
     - arcv.vwscnmsac.vs (changed from .vx to .vs, scalar is LMUL=1)
     - arcv.vwscjmac.vs (changed from .vx to .vs, scalar is LMUL=1)
     - arcv.vwscjnmsac.vs (changed from .vx to .vs, scalar is LMUL=1)

    New argument type configurations (wvl1s_ops, wwl1v_ops) and operand infos
    (i_wvl1s_ops, i_wwl1v_ops) are introduced to properly handle the
    double-truncated LMUL=1 scalar operands.  Separate UNSPEC constants distinguish
    .vs scalar variants from .vv vector-vector forms.  Overloading is disabled as
    these operations require explicit type specifications.

    Signed-off-by: Luis Silva <luiss@synopsys.com>


luismgsilva:
Different approaches were attempted, but none were successful.
New UNSPEC_*_SCALAR codes were introduced so that GCC can distinguish between the .vv and .vs variants; otherwise, the .vv form is always emitted.

An attempt was made to use a single pattern and select between .vv and .vs based on the mode, but .vv was consistently chosen.

gcc/config/riscv/arcv-vector.md
```diff
@@ -1910,11 +1918,11 @@
 	(unspec:VWEXTI
 	[(sign_extend:VWEXTI
 	  (match_operand:<V_DOUBLE_TRUNC> 3 "register_operand" "vr,vr,vr,vr,vr,vr,vr,vr,vr,vr,vr,vr"))
-	(match_operand:<VSUBEL> 4 "register_operand" "r,r,r,r,r,r,r,r,r,r,r,r")]
-	  UNSPEC_ARCV_VWSCMUL)
+	(match_operand:<V_DOUBLE_TRUNC> 4 "register_operand" "vr,vr,vr,vr,vr,vr,vr,vr,vr,vr,vr,vr")]
```
MichielDerhaeg: Why not LMUL1?

MichielDerhaeg:
> Different approaches were attempted, but none were successful. New UNSPEC_*_SCALAR codes were introduced so that GCC can distinguish between the .vv and .vs variants; otherwise, the .vv form is always emitted.
> 
> An attempt was made to use a single pattern and select between .vv and .vs based on the mode, but .vv was consistently chosen.

I think your approach makes sense. For the LMUL1 case, these will otherwise overlap. I think this is the only way.

gcc/config/riscv/arcv-vector.md
MichielDerhaeg: Needs earlyclobber

MichielDerhaeg, line 1815: Needs earlyclobber, source and destination registers are not allowed to overlap.

MichielDerhaeg (changes requested):

gcc/config/riscv/arcv-vector.md
luismgsilva: To my understanding the argument type would need to change from `RVV_BASE_double_trunc_vector` to something like `RVV_BASE_double_trunc_lmul1_vector` which currently doesnt exist. Not sure if implementing it would be the right approach.

gcc/testsuite/gcc.target/riscv/arcv-vcplx-vscjmul_vs-compile-1.c
```diff
@@ -0,0 +1,60 @@
...
+**   csrwi\s+vxrm,0
+**   vsetvli\s+zero,\s*[a-x0-9]+,\s*e16,m1,\s*t[au],\s*m[au]
+**   (?:vmv[0-9]*r\.v\s+(?:v[0-9]|v[1-2][0-9]|v3[0-1]),\s*(?:v[0-9]|v[1-2][0-9]|v3[0-1])\n\s+)+arcv.vscjmul.vs\s+(?:v[0-9]|v[1-2][0-9]|v3[0-1]),\s*(?:v[0-9]|v[1-2][0-9]|v3[0-1]),\s*[a-x0-9]+
+**   ret
+*/
+vint16m1_t
+test_vscjmul_vs_i16 (vint16m1_t vs2, vint16m1_t vs1, size_t vl)
+{
+  return __riscv_arcv_vscjmul_vs_i16m1 (vs2, vs1, 0, vl);
+}
```
MichielDerhaeg, line 21: You're testing any case other than LMUL1. So we don't verify that the right vector parameter stays on LMUL1. We need to add lmul > 1 cases.

gcc/config/riscv/arcv-vector.md
MichielDerhaeg: Yes, we have to because for lmul2 we now need both vector parameters to have the same type, which is not correct:
```lang=c
vint32m4_t
test_vwscjmul_vs_i16m2 (vint16m2_t vs2, vint16m2_t vs1, size_t vl)
{
  return __riscv_arcv_vwscjmul_vs_i32m4 (vs2, vs1, vl);
}

```

MichielDerhaeg (changes requested):

MichielDerhaeg (approved):

luismgsilva:
@MichielDerhaeg please see https://github.com/foss-for-synopsys-dwc-arc-processors/binutils-gdb/pull/100/

PR #208, opened 2026-02-12
arcv: Fix register overlap case for MAC fusion

MichielDerhaeg:
The MAC fusion requires a specific order in the input operands of add.


MichielDerhaeg:
This will likely be temporary as the case with the trailing mv is very suboptimal and we will remove it.

PR #209, opened 2026-02-12
arcv: XBFU fusion for RPX

MichielDerhaeg:
Extend the zero_extract_fused pattern to work for 64-bit targets.


luismgsilva (approved):
Looks good to me. Thanks !

gcc/testsuite/gcc.target/riscv/arcv-fusion-xbfu.c
```diff
@@ -1,14 +1,24 @@
 /* { dg-do compile } */
-/* { dg-require-effective-target rv32 } */
 /* { dg-skip-if "" { *-*-* } { "-g" "-flto" "-O0" "-Oz" "-Os" } } */
-/* { dg-options "-mtune=arc-v-rhx-100-series -march=rv32im_zbs -mabi=ilp32" } */
+/* { dg-options "-mtune=arc-v-rhx-100-series -march=rv32im_zbs -mabi=ilp32" { target { rv32 } } } */
+/* { dg-options "-mtune=arc-v-rhx-100-series -march=rv64im_zbs -mabi=lp64" { target { rv64 } } } */
```
MichielDerhaeg: This is probably not what we want. I need to check if this tests correctly on multilib toolchains.

luismgsilva: For multilib, we might need to separate this test case into two files, 32 and 64, so that it will test for both archs.

PR #212, opened 2026-02-23
arcv: Fix operand overlaps for vnsra intrinsic RTL patterns.

luismgsilva:
The vnsra intrinsics with .s and .2s suffixes were incorrectly allowing the destination register to match operand 4 (the shift amount) in their RTL patterns.  This created illegal register overlap alternatives in the constraint set.

This patch removes the problematic constraint alternatives where operand 4 was constrained to "0" (matching the destination register).

The following RTL patterns are corrected:
  @pred_narrow_arcv_vnsra_s<mode> (arcv.vnsra.s.w)
  @pred_quad_narrow_arcv_vnsra_s<mode> (arcv.vnsra.s.q)
  @pred_narrow_arcv_vnsra_2s<mode> (arcv.vnsra.2s.w)
  @pred_quad_narrow_arcv_vnsra_2s<mode> (arcv.vnsra.2s.q)

Each pattern now has 8 alternatives instead of 12, eliminating the constraint variants where the shift operand incorrectly matched the destination.

### DIFF
--- a/gcc/config/riscv/arcv-vector.md
+++ b/gcc/config/riscv/arcv-vector.md
@@ -565,50 +565,50 @@
    (set_attr "mode" "<V_QUAD_TRUNC>")])
 
 (define_insn "@pred_narrow_arcv_vnsra_s<mode>"
-  [(set (match_operand:<V_DOUBLE_TRUNC> 0 "register_operand" "=&vd,&vd,&vr,&vr,&vd,&vd,&vr,&vr,&vd,&vd,&vr,&vr")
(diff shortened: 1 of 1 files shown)


luismgsilva:
@MichielDerhaeg This patch fixes the extra failures from the unmerged tests mentioned in the meeting. Not sure if we should consider them before the tests are merged.

MichielDerhaeg (approved):

MichielDerhaeg:
> @MichielDerhaeg This patch fixes the extra failures from the unmerged tests mentioned in the meeting. Not sure if we should consider them before the tests are merged.

I think it is fine to merge, looks more correct than before.

PR #214, opened 2026-02-25
arcv: Fix register constraints for XARCVvcplx vscjmul intrinsic.

luismgsilva:
The vscjmul pattern incorrectly allowed the destination register to overlap with
source operands, causing register corruption when alternatives 0-3 were
selected.

This patch corrects the RTL pattern to:
  - Add early-clobber (&) to operand 0 for alternatives 0-3 to prevent overlap
    with operand 3.
  - Tie operand 4 to operand 0 (constraint "0") for alternatives 0-3 to enable
    proper two-operand form.

Alternatives 4-11 remain unchanged as three-operand forms where all operands are
independent

### DIFF
--- a/gcc/config/riscv/arcv-vector.md
+++ b/gcc/config/riscv/arcv-vector.md
@@ -1836,7 +1836,7 @@
    (set_attr "mode" "<MODE>")])
 
 (define_insn "@pred_arcv_vscjmul<mode>"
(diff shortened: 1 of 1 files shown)


luismgsilva:
Closing this PR as this is not a real problem but rather an issue on internal testing.

PR #215, opened 2026-02-25
Add lint checks on pull requests

exur00:


exur00:
The results can be seen in https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/233 and https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/234

.github/workflows/gcc-contrib-lint.yml
```diff
@@ -0,0 +1,166 @@
...
+          LIBDIR=$(dirname $(ldconfig -p | grep libgdiagnostics.so.0 | awk '{print $NF}' | head -1))
+          LIBFILE=$(basename $(ldconfig -p | grep libgdiagnostics.so.0 | awk '{print $NF}' | head -1))
+          sudo ln -sf "$LIBFILE" "$LIBDIR/libgdiagnostics.so"
+          python3 -m pip install --user unidiff termcolor
+
+      - name: Fetch GCC contrib scripts
+        run: |
+          mkdir -p gcc-contrib
+          curl -sL -o gcc-contrib/check_GNU_style.py \
+            https://raw.githubusercontent.com/gcc-mirror/gcc/master/contrib/check_GNU_style.py
```
luismgsilva: Not sure how I feel about using gcc-mirror for this. Should probably be fine but we can't be sure what is pushed to it. Or even upstream gcc for this matter.

luismgsilva (approved):

.github/workflows/gcc-contrib-lint.yml
exur00: I just realised those files are in this repo, I can just use them from here.

MichielDerhaeg:
Is this still blocked by something?

PR #218, opened 2026-03-02
arcv: fix/refactor for fusion for rmx500/rhx100/rpx100

Alexehv77:
Take care that fusion patters for ld/st followed/preceded by op/op-imm follow the spec. In the comments you can see to which fusion pattern each condition belongs to.
Bechmarked rhx100 and rmx500 on coremark and found out that only post-increments are being triggered.

### DIFF
--- a/gcc/config/riscv/arcv.cc
+++ b/gcc/config/riscv/arcv.cc
@@ -180,125 +180,78 @@ arcv_arith_type_insn_p (rtx_insn *insn)
 	 || type == TYPE_CTZ);
 }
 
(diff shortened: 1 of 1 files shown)


Alexehv77:
The issue with the build was that needed to filter out instructions with immediate single source operand (as they cant be fused with):
https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/blob/e367dfa3864c0459a4482e60a3d6cc0165ad8e70/gcc/config/riscv/arcv.cc#L207

luismgsilva (approved):

MichielDerhaeg (approved):

PR #219, opened 2026-03-02
arcv: Fix 64-bit scalar operand handling for vsra, vsrat, and vsra_2s…

luismgsilva:
… on RV32.

The vsra, vsrat, and vsra_2s scalar intrinsics were not properly handling 64-bit scalar register operands (DImode) on 32-bit systems (RV32).  The scalar operand needs to be split into two 32-bit registers via sew64_scalar_helper, but the patterns were missing the necessary logic.

This patch adds define_expand patterns for V_VLSI_D modes (64-bit elements) that invoke sew64_scalar_helper to properly split DImode scalar operands into register pairs.  The V_VLSI_QHS patterns (8/16/32-bit elements) remain as define_insn since they don't require special handling.


luismgsilva:
This patch has been approved offline.

PR #220, opened 2026-03-03
arcv: fix for fused mul+add followed by mv

MichielDerhaeg:
the fix should impact cormark for rpx100

The patch refers to the case where due to a fused
mul+add there is need for an extra move operation
that is being placed at the end of the 2 operations. I prohibit the case where such move would be placed after the add move present at the end of the
2 instructions. In return sometimes there can be a move place at the beginning of the sequence.

jira P10019563-84761

### DIFF
--- a/gcc/config/riscv/riscv.md
+++ b/gcc/config/riscv/riscv.md
@@ -4504,46 +4504,32 @@
 )
 
 (define_insn "madd_split_fused"
-  [(set (match_operand:SI 0 "register_operand" "=&r,r")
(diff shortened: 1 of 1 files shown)


MichielDerhaeg:
@Alexehv77 Can you please address the remarks I made on the Jira?

Alexehv77:
> P10019563-84761

I removed the 2 original patterns and i provided in the jira ticket in-depth explanations including the benchmarking results for rpx100. 
https://jira.internal.synopsys.com/browse/P10019563-84761

PR #222, opened 2026-03-13
arcv: Backport RVA23 support

MichielDerhaeg:
These extensions don't enable any instructions or functionality. They are just "marker" extension that only inform the assembler that they are enabled.


kolerov:
@MichielDerhaeg Do the similar patches exist in upstream?

MichielDerhaeg:
> @MichielDerhaeg Do the similar patches exist in upstream?

Upstream has this functionality but they refactored the riscv extension system before most of these were added so cherry-picking is not an option.
The dejagnu tests yield the same results for GCC 16.

kolerov (approved):

luismgsilva (approved):

PR #224, opened 2026-03-18
niter: Prove IV's with non-constant steps can't overflow

MichielDerhaeg:
This is step one of loop versioning for overflowing IV's.
This allows GCC to prove that the IV doesn't overflow when a bounds check is present in the source.
More specifically, IV's that have non-constant steps, like the indices used in the matrix operation in the dejagnu test.


MichielDerhaeg:
Note to self: I need to verify the test works for rv32

MichielDerhaeg:
This was also bootstrapped and regtested on x86_64.

luismgsilva (approved):

PR #226, opened 2026-03-20
ivbound: Introduce loop versioning pass for potentially overflowing IV's

MichielDerhaeg:
Finds loops with potentially overflowing IV's and versions the loop with a constant bound such that the bounded loop can have improved optimizations.


MichielDerhaeg:
I'm going to merge it, but feel free to review it post commit.

PR #228, opened 2026-03-23
riscv: rpx100: emit mulh(s/u) mul pairs in that order

exur00:
RPX100 allows a mulh(s/u) followed by a mul operation to be bonded, saving 3 cycles in the multiplier and allowing the mul's result to be used 1 cycle earlier than expected.

- Add a mul_part attribute to al multiplication instructions in riscv.md to distinguish instructions that return the high and low part.
- Define_expand <us>mulditi3 is modified to emit the high part first for rpx100
- To preserve this order, a bonded define_insn_and_split is introduced
- The same things for the usmul... variants
- Detect this case in riscv_macro_fusion_pair_p
- riscv_sched_adjust_cost reflects the mul result being available 1 cycle sooner in this case

TODO: the cprop_hardreg1 pass sometimes breaks the matching input constraint for the bonded pair, preventing the bonded execution.

### DIFF
--- a/gcc/config/riscv/arcv-rpx100.md
+++ b/gcc/config/riscv/arcv-rpx100.md
@@ -24,7 +24,7 @@
 (define_cpu_unit "arcv_rpx100_ALU_A_fuse1_early"	"arcv_rpx100")
 (define_cpu_unit "arcv_rpx100_ALU_B_fuse0_early"	"arcv_rpx100")
 (define_cpu_unit "arcv_rpx100_ALU_B_fuse1_early"	"arcv_rpx100")
-(define_cpu_unit "arcv_rpx100_MPY32"	"arcv_rpx100")
+(define_cpu_unit "arcv_rpx100_MPY"	"arcv_rpx100")
 (define_cpu_unit "arcv_rpx100_DIV"	"arcv_rpx100")
 (define_cpu_unit "arcv_rpx100_DMP_fuse0"	"arcv_rpx100")
 (define_cpu_unit "arcv_rpx100_DMP_fuse1"	"arcv_rpx100")
@@ -45,7 +45,7 @@
 (define_insn_reservation "arcv_rpx100_imul_fused" 4
   (and (eq_attr "tune" "arcv_rpx100")
        (eq_attr "type" "imul_fused"))
-  "(arcv_rpx100_issueA_fuse0 + arcv_rpx100_issueA_fuse1 + arcv_rpx100_ALU_A_fuse0_early + arcv_rpx100_ALU_A_fuse1_early + arcv_rpx100_MPY32), nothing*3")
+  "(arcv_rpx100_issueA_fuse0 + arcv_rpx100_issueA_fuse1 + arcv_rpx100_ALU_A_fuse0_early + arcv_rpx100_ALU_A_fuse1_early + arcv_rpx100_MPY), nothing*3")
 
 (define_insn_reservation "arcv_rpx100_alu_fused" 1
    (and (eq_attr "tune" "arcv_rpx100")
@@ -64,8 +64,26 @@
 
 (define_insn_reservation "arcv_rpx100_mpy32_insn" 4
   (and (eq_attr "tune" "arcv_rpx100")
-       (eq_attr "type" "imul"))
-  "arcv_rpx100_issueA_fuse0 + arcv_rpx100_MPY32, nothing*3")
+       (eq_attr "type" "imul")
+       (eq_attr "mode" "SI"))
+  "arcv_rpx100_issueA_fuse0 + arcv_rpx100_MPY, nothing*3")
+
+(define_insn_reservation "arcv_rpx100_mpy64l_insn" 6
+  (and (eq_attr "tune" "arcv_rpx100")
+       (eq_attr "type" "imul")
+       (eq_attr "mode" "DI")
+       (eq_attr "mul_part" "low"))
+  "arcv_rpx100_issueA_fuse0 + arcv_rpx100_MPY, arcv_rpx100_MPY*2, nothing * 3")
+
+;; The 1 cycle reservation of the multiplier is only correct for the bonded mul case.
+;; Mul high part is usually generated together with the low part.
+;; Otherwise, this is 3 cycles too optimistic about multiplier reservation.
+(define_insn_reservation "arcv_rpx100_mpy64h_insn" 7
+  (and (eq_attr "tune" "arcv_rpx100")
+       (eq_attr "type" "imul")
+       (eq_attr "mode" "DI")
+       (eq_attr "mul_part" "high"))
+  "arcv_rpx100_issueA_fuse0 + arcv_rpx100_MPY, nothing*6")
 
 (define_insn_reservation "arcv_rpx100_load_insn" 3
   (and (eq_attr "tune" "arcv_rpx100")
--- a/gcc/config/riscv/arcv.cc
+++ b/gcc/config/riscv/arcv.cc
@@ -336,6 +336,79 @@ arcv_memop_lui_pair_p (rtx_insn *prev, rtx_insn *curr)
   return false;
 }
 
+/* Check if RTX matches multiply-high pattern:
+   (set (reg:DI DEST)
+    (truncate:DI (lshiftrt:TI (mult:TI (extend:TI (OP0))
+					(extend:TI (OP1)))
+				(const_int 64)))  */
+
+static bool
+is_multiply_high_pattern_p (rtx src)
+{
+  if (GET_CODE (src) != SET)
+    return false;
+  rtx truncate = SET_SRC (src);
+
+  if (GET_CODE (truncate) != TRUNCATE)
+    return false;
+
+  rtx lshiftrt = XEXP (truncate, 0);
+  if (GET_CODE (lshiftrt) != LSHIFTRT)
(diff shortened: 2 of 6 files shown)


gcc/config/riscv/riscv.cc
```diff
@@ -10334,7 +10334,7 @@ riscv_macro_fusion_pair_p (rtx_insn *prev, rtx_insn *curr)
     }
 
   if (TARGET_ARCV_FUSION)
-    return arcv_macro_fusion_pair_p (prev, curr);
+    return arcv_macro_fusion_pair_p (prev, curr) || arcv_macro_bonded_mul_pair_p (prev, curr);
```
MichielDerhaeg: Line is too long for GNU code style?

gcc/config/riscv/arcv.cc
```diff
@@ -336,6 +336,75 @@ arcv_memop_lui_pair_p (rtx_insn *prev, rtx_insn *curr)
...
 }
 
+/* Check if RTX matches multiply-high pattern:
+   (set (reg:DI DEST)
+    (truncate:DI (lshiftrt:TI (mult:TI (extend:TI (OP0))
+					(extend:TI (OP1)))
+				(const_int 64)))  */
+
+static bool
+is_multiply_high_pattern (rtx src)
```
MichielDerhaeg:
```suggestion
is_multiply_high_pattern_p (rtx src)
```
Can this predicate be used instead of the attribute you attach to low and high part instructions?

gcc/config/riscv/arcv-rpx100.md
```diff
@@ -64,8 +64,23 @@
...
+  "arcv_rpx100_issueA_fuse0 + arcv_rpx100_MPY, nothing*3")
+
+(define_insn_reservation "arcv_rpx100_mpy64l_insn" 6
+  (and (eq_attr "tune" "arcv_rpx100")
+       (eq_attr "type" "imul")
+       (eq_attr "mode" "DI")
+       (eq_attr "mul_part" "low"))
+  "arcv_rpx100_issueA_fuse0 + arcv_rpx100_MPY, arcv_rpx100_MPY*2, nothing * 3")
+
+(define_insn_reservation "arcv_rpx100_mpy64h_insn" 7
```
MichielDerhaeg, line 81: You make some assumptions here, and you should document them.

gcc/config/riscv/arcv.cc
```diff
@@ -730,11 +841,40 @@ arcv_sched_adjust_priority (rtx_insn *insn, int priority)
...
 int
-arcv_sched_adjust_cost (rtx_insn *insn, int dep_type, int cost)
+arcv_sched_adjust_cost (rtx_insn *insn, int dep_type, rtx_insn *dep_insn, int cost)
 {
+  /* Check for bonded 64-bit multiply pair on arcv_rpx100.  */
+  if (riscv_microarchitecture == arcv_rpx100
+      && arcv_rpx100_depends_on_lowpart_bonded_mul (insn, dep_type, dep_insn))
+    {
+      return cost - 1;
+    }
```
MichielDerhaeg:
```suggestion
  if (riscv_microarchitecture == arcv_rpx100
      && arcv_rpx100_depends_on_lowpart_bonded_mul (insn, dep_type, dep_insn))
    return cost - 1;
```

gcc/config/riscv/arcv.cc
```diff
@@ -730,11 +841,40 @@ arcv_sched_adjust_priority (rtx_insn *insn, int priority)
...
+
+  if (!arcv_macro_bonded_mul_pair_p (prev_insn, dep_insn))
+    return false;
+
+  if (dump_file)
+    {
+      fprintf (dump_file, "ARCV_BONDED_MUL_ADJUST_COST\t insn: %d, high_part: %d, low_part: %d\n",
+		INSN_UID (insn), INSN_UID (prev_insn), INSN_UID (dep_insn));
+    }
+
```
MichielDerhaeg, line 865:
```suggestion
  if (dump_file)

    fprintf (dump_file, "ARCV_BONDED_MUL_ADJUST_COST\t insn: %d, high_part: %d, low_part: %d\n",
	  INSN_UID (insn), INSN_UID (prev_insn), INSN_UID (dep_insn));

```

MichielDerhaeg (changes requested):
You can add a few tests that verify sched2 output

gcc/config/riscv/arcv.cc
exur00: I think it cannot

gcc/config/riscv/riscv.md
```diff
@@ -1128,7 +1132,8 @@
   "TARGET_HARD_FLOAT  || TARGET_ZFINX"
   "fmul.<fmt>\t%0,%1,%2"
   [(set_attr "type" "fmul")
-   (set_attr "mode" "<UNITMODE>")])
+   (set_attr "mode" "<UNITMODE>")
+   (set_attr "mul_part" "low")])
```
MichielDerhaeg: Are these needed? there is a default

gcc/config/riscv/riscv.md
```diff
@@ -1294,16 +1303,56 @@
...
   rtx high = gen_reg_rtx (DImode);
-  emit_insn (gen_<su>muldi3_highpart (high, operands[1], operands[2]));
+
+  if (riscv_microarchitecture == arcv_rpx100)
+    {
+      emit_insn (gen_<su>mulditi4_bonded_arcv_rpx100 (high,
+      						      operands[1],
+						      operands[2],
+						      low));
+    }
```
MichielDerhaeg:
```suggestion
      emit_insn (gen_<su>mulditi4_bonded_arcv_rpx100 (high,
      						      operands[1],
						      operands[2],
						      low));
```
Formatting looks a bit odd here

gcc/config/riscv/riscv.cc
```diff
@@ -10334,7 +10334,8 @@ riscv_macro_fusion_pair_p (rtx_insn *prev, rtx_insn *curr)
     }
 
   if (TARGET_ARCV_FUSION)
-    return arcv_macro_fusion_pair_p (prev, curr);
+    return (arcv_macro_fusion_pair_p (prev, curr)
+	    || arcv_macro_bonded_mul_pair_p (prev, curr));
```
MichielDerhaeg: You should check for RPX.

gcc/testsuite/gcc.target/riscv/arcv-rpx100-bonded-mul-1.c
```diff
@@ -0,0 +1,30 @@
...
+__uint128_t
+baz (uint64_t x, uint64_t y)
+{
+  __uint128_t z = (__uint128_t) x * y;
+  return z + 1;
+}
+
+/* { dg-final { scan-rtl-dump "0-->.*a4=trn\\(sxn\\(a0\\)\\*sxn\\(a1\\) 0>>0x40\\).*\n.*1-->.*a5=a0\\*a1.*\n.*6-->.*a0=a5\\+0x1.*" "sched2" } } */
+/* { dg-final { scan-rtl-dump "0-->.*a4=trn\\(zxn\\(a1\\)\\*sxn\\(a0\\) 0>>0x40\\).*\n.*1-->.*a5=a1\\*a0.*\n.*6-->.*a0=a5\\+0x1.*" "sched2" } } */
+/* { dg-final { scan-rtl-dump "0-->.*a4=trn\\(zxn\\(a0\\)\\*zxn\\(a1\\) 0>>0x40\\).*\n.*1-->.*a5=a0\\*a1.*\n.*6-->.*a0=a5\\+0x1.*" "sched2" } } */
```
MichielDerhaeg, line 30: You didn't adjust the variable_issue or the reorder2 hook. Do we handle fusion with the bonded instructions correctly?

exur00: variable_issue: I don't think this is needed, because it calculates how many insns can be emitted this 1 cycle, but the bonded mul is issued in over 2 cycles, so no changes.
fusion: only mulw has fusions, not mul or mulh(s/u), so no changes.
reorder2: do you think this is necessary? with sched_adjust_cost and macro_fusion_pair_p changes gcc seems to put them together and keep them together without intervention? As the bonded pair won't fuse with anything, and the adjust cost informs gcc that putting them together in that order makes it cheaper, I would think GCC can already take care of this as is.

gcc/config/riscv/riscv.md
exur00: I removed the braces, it still looks a bit weird, but aligning the argument does seem to be the usual way (see line 2366, 3648)

gcc/config/riscv/riscv.cc
```diff
@@ -10335,7 +10335,8 @@ riscv_macro_fusion_pair_p (rtx_insn *prev, rtx_insn *curr)
 
   if (TARGET_ARCV_FUSION)
     return (arcv_macro_fusion_pair_p (prev, curr)
-	    || arcv_macro_bonded_mul_pair_p (prev, curr));
+	    || (riscv_microarchitecture == rpx100 &&
+		arcv_macro_bonded_mul_pair_p (prev, curr)));
```
MichielDerhaeg: I would put that in a separate `if`

MichielDerhaeg (approved):

PR #229, opened 2026-03-27
target dependent phase to improves memory accesses

Alexehv77:
This phase identifies memory loads that use a base address calculated via instructions like sh1add. It traces the index register of that calculation back through previous instructions to find increments.
As a result instead of having a separate instruction to increment followed by a pointer to scale and add it, this pass merges the constant offset directly into the memory load. Thus it reduces the total number of instructions leading to smaller code size and faster execution.

### DIFF
--- a/gcc/config/riscv/riscv-fold-mem.cc
+++ b/gcc/config/riscv/riscv-fold-mem.cc
@@ -1,0 +1,356 @@
+/* RISC-V memory folding pass
+ * The phase is designed to exploit the zba extensions by targeting
+ * address generation patterns where an index is being incremented
+ * and then used to access an array.  The phase folds constants into
+ * memory accesses: instead of performing a separate add and shift
+ * or shiftadd followed by a load it folds the cumulative increment
+ * directly into the offset field of the load/store instruction.
+ * As a result it replaces two or more instructions (an address
+ * calculation and a memory access) with a single instruction which
+ * reduces both code size and execution time.  */
+
+#include "config.h"
+#include "system.h"
+#include "coretypes.h"
+#include "backend.h"
+#include "rtl.h"
+#include "df.h"
+#include "options.h"
+#include "insn-config.h"
+#include "recog.h"
+#include "target.h"
+#include "tree-pass.h"
+#include "memmodel.h"
+#include "emit-rtl.h"
+#include "rtl-iter.h"
+#include <vector>
+
+namespace
+{
+
+/* Configuration data.  */
+const pass_data pass_data_riscv_fold_mem = {
+  RTL_PASS,		/* type.  */
+  "riscv_fold_mem",	/* name.  */
+  OPTGROUP_LOOP,	/* optinfo_flags.  */
+  TV_NONE,
+  0,
+  0,
+  0,
+  0,
+  0,
+};
+
+class pass_riscv_fold_mem : public rtl_opt_pass
+{
+public:
+  pass_riscv_fold_mem (gcc::context *ctxt)
+      : rtl_opt_pass (pass_data_riscv_fold_mem, ctxt)
+  {
+  }
+
+  /* run if zba extension and 64 bit mode are enabled.  */
+  virtual bool
+  gate (function *) final override
+  {
+    return optimize > 0 && TARGET_64BIT && riscv_fold_mem;
+  }
+
+  virtual unsigned int execute (function *) override;
+
+private:
+  rtx_insn *find_def (rtx_insn *insn, rtx reg);
+  rtx_insn *find_nearest_def_in_bb (rtx_insn *insn, rtx reg);
+  rtx trace_increments (rtx_insn *insn, rtx reg, HOST_WIDE_INT *total_offset);
+};
+
+/* Find the unique definition of REG used in INSN.  */
+rtx_insn *
+pass_riscv_fold_mem::find_def (rtx_insn *insn, rtx reg)
+{
+  if (!reg || !REG_P (reg))
+    return NULL;
+
+  df_ref use = df_find_use (insn, reg);
+  if (!use)
+    return NULL;
+
+  struct df_link *defs = DF_REF_CHAIN (use);
+  if (!defs || defs->next)
+    return NULL;
+
+  return DF_REF_INSN (defs->ref);
+}
+
+/* Locate the definition of REG that most closely precedes INSN
+   within the same basic block.  */
+rtx_insn *
+pass_riscv_fold_mem::find_nearest_def_in_bb (rtx_insn *insn, rtx reg)
+{
+  df_ref def_ref;
+  rtx_insn *best_def = NULL;
+  int current_luid = DF_INSN_LUID (insn);
+  int best_luid = -1;
+  basic_block curr_bb = BLOCK_FOR_INSN (insn);
+
+  for (def_ref = DF_REG_DEF_CHAIN (REGNO (reg)); def_ref;
+       def_ref = DF_REF_NEXT_REG (def_ref))
+    {
+      rtx_insn *def_insn = DF_REF_INSN (def_ref);
+
+      if (BLOCK_FOR_INSN (def_insn) == curr_bb)
+	{
+	  int def_luid = DF_INSN_LUID (def_insn);
+
+	  if (def_luid < current_luid && def_luid > best_luid)
+	    {
+	      best_luid = def_luid;
+	      best_def = def_insn;
+	    }
+	}
+    }
+  return best_def;
+}
+
+/* Recursively trace register increments to find the root register
+ * and calculate the cumulative offset.  */
+rtx
+pass_riscv_fold_mem::trace_increments (rtx_insn *insn, rtx reg,
+				       HOST_WIDE_INT *total_offset)
+{
+  if (!reg || !REG_P (reg))
+    return reg;
+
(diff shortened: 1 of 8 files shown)


gcc/config/riscv/riscv-fold-mem.cc
```diff
@@ -0,0 +1,219 @@
...
+        df_chain_add_problem(DF_DU_CHAIN | DF_UD_CHAIN);
+        df_analyze();
+
+        rtx_insn *insn;
+        for (insn = get_insns(); insn; insn = NEXT_INSN(insn)) {
+            if (!NONDEBUG_INSN_P(insn)) continue;
+
+            rtx set_load = single_set(insn);
+            if (!set_load || (!MEM_P(SET_SRC(set_load)) && 
+                              GET_CODE(SET_SRC(set_load)) != SIGN_EXTEND))
```
MichielDerhaeg: I don't think we should be limited to loads. I think stores should also apply.

MichielDerhaeg: `SIGN_EXTEND` should be optional? We could be loading register size values?

gcc/config/riscv/riscv-fold-mem.cc
```diff
@@ -0,0 +1,219 @@
...
+
+            df_ref addr_use = df_find_use(insn, addr_reg);
+            if (!REG_P(addr_reg) || !addr_use || 
+                !DF_REF_CHAIN(addr_use) || DF_REF_CHAIN(addr_use)->next) 
+		    continue;
+
+            rtx_insn *insn_addr = DF_REF_INSN(DF_REF_CHAIN(addr_use)->ref);
+
+            rtx set_addr = single_set(insn_addr);
+            if (!set_addr || GET_CODE(SET_SRC(set_addr)) != PLUS) continue;
```
MichielDerhaeg: Whenever you try to match RTL structures, adding the RTL in the comments could really help with understanding.

gcc/config/riscv/riscv.cc
```diff
@@ -4031,7 +4031,7 @@ riscv_rtx_costs (rtx x, machine_mode mode, int outer_code, int opno ATTRIBUTE_UN
 	      && !riscv_compressed_lw_address_p (XEXP (x, 0)))
 	    cost++;
 
-	  *total = COSTS_N_INSNS (cost + tune_param->memory_cost);
```
MichielDerhaeg, line 4034: what's this for?

gcc/config/riscv/riscv-fold-mem.cc
```diff
@@ -0,0 +1,219 @@
...
+        rtx set = single_set(def_insn);
+        if (!set) return reg;
+
+        rtx src = SET_SRC(set);
+        // Look for: (sign_extend (plus (subreg (reg)) (const_int)))
+        if (GET_CODE(src) == SIGN_EXTEND && GET_CODE(XEXP(src, 0)) == PLUS) {
+            rtx plus_op = XEXP(src, 0);
+            if (CONST_INT_P(XEXP(plus_op, 1))) {
+                *total_offset += INTVAL(XEXP(plus_op, 1));
+                rtx inner_reg = XEXP(XEXP(plus_op, 0), 0); // Unwrap subreg
```
MichielDerhaeg: Should you check for subreg?

gcc/config/riscv/riscv-fold-mem.cc
```diff
@@ -0,0 +1,219 @@
...
+const pass_data pass_data_riscv_fold_mem = {
+  RTL_PASS, "riscv_fold_mem", OPTGROUP_LOOP, TV_NONE, 0, 0, 0, 0, 0,
+};
+
+class pass_riscv_fold_mem : public rtl_opt_pass {
+public:
+  pass_riscv_fold_mem (gcc::context *ctxt) : rtl_opt_pass (pass_data_riscv_fold_mem, ctxt) {}
+  
+  virtual bool gate (function *) final override { 
+    return TARGET_64BIT && TARGET_ZBA && optimize >= 2;
```
MichielDerhaeg: I would not limit this to 64-bit if possible.
What kind of code is generated when zba is not used?

gcc/config/riscv/riscv-fold-mem.cc
```diff
@@ -0,0 +1,219 @@
...
+
+            if (!MEM_P(mem)) 
+		    continue;
+
+            rtx addr_reg = XEXP(mem, 0);
+	    if (!REG_P(addr_reg)) 
+		    continue;
+
+            df_ref addr_use = df_find_use(insn, addr_reg);
+            if (!REG_P(addr_reg) || !addr_use ||
```
MichielDerhaeg: You already checked REG_P before.

gcc/config/riscv/riscv-fold-mem.cc
```diff
@@ -0,0 +1,219 @@
...
+
+            HOST_WIDE_INT cumulative_inc = 0;
+            rtx root_reg = trace_increments(insn_addr, index_reg, &cumulative_inc);
+
+            if (cumulative_inc == 0 || root_reg == index_reg) continue;
+
+            HOST_WIDE_INT scaled_offset = cumulative_inc << shift_amount;
+            confirm_change_group();
+
+            validate_change(insn_addr, &XEXP(XEXP(and_rtx, 0), 0), root_reg, 1);
```
MichielDerhaeg: Why are you changing insn_addr, can't you use root_reg directly in new_mem_addr?
You also don't check if insn_addr is only used by insn, so you now break other uses.

gcc/config/riscv/riscv-fold-mem.cc
```diff
@@ -0,0 +1,219 @@
...
+            rtx new_mem_addr = plus_constant(Pmode, addr_reg, scaled_offset);
+            validate_change(insn, &XEXP(mem, 0), new_mem_addr, 1);
+
+            if (apply_change_group()) {
+                df_insn_rescan_all();
+            }
+	}
+
+        basic_block bb;
+        FOR_EACH_BB_FN(bb, fun) {
```
MichielDerhaeg: What's the point of this? Can this be handled by cse or cprop passes instead?

gcc/config/riscv/riscv.cc
```diff
@@ -11162,8 +11181,54 @@ riscv_option_override (void)
      while processing functions with potential target attributes.  */
   target_option_default_node = target_option_current_node
     = build_target_option_node (&global_options, &global_options_set);
+
+  riscv_register_pass(make_pass_riscv_fold_mem, PASS_POS_INSERT_AFTER, "combine");
```
MichielDerhaeg: shouldn't this go in riscv-passes.def?

MichielDerhaeg (changes requested):

gcc/config/riscv/riscv-fold-mem.cc
Alexehv77: i allowed also stores

Alexehv77: i made it optional so it works also for memory access rtx-es that arent sign extended.

gcc/config/riscv/riscv-fold-mem.cc
Alexehv77: i added explanation see lines 200-201

gcc/config/riscv/riscv.cc
Alexehv77: i changed it back - was from prior code where i tried to change change instruction costs

gcc/config/riscv/riscv-fold-mem.cc
Alexehv77: it is checked - see line 145

gcc/config/riscv/riscv-fold-mem.cc
Alexehv77: the optimization was driven by the performance loss observed in riscv64 with zba. it took into account the shiftadds together with immediate adds and loads. I only checked correctness for when zba is not used. I can try dive into perfromance when zba is not used. I assume that shitadds are replaced by separate add and shift.

gcc/config/riscv/riscv-fold-mem.cc
Alexehv77: code was refactored by prior commit

gcc/config/riscv/riscv-fold-mem.cc
Alexehv77: the insn_addr needs to pick the root of the chain of immediate additions so needs to be changed.
the output of insn_addr is not changed. I only change one of its inputs and the accumulation of the additions i pass it directly into the offset of the true dependent subsequent memory operations.

gcc/config/riscv/riscv-fold-mem.cc
Alexehv77: i tried placing the phase in various places where there were not available those after.

gcc/config/riscv/riscv.cc
Alexehv77: i moved it there.

gcc/config/riscv/riscv.cc
MichielDerhaeg: It's still there

Alexehv77: check now

PR #231, opened 2026-04-03
RISC-V: Omit ghost from the pipeline-checker output

MichielDerhaeg:


MichielDerhaeg:
Accepted upstream, no need for it downstream

PR #232, opened 2026-04-03
RISC-V: Move fusion-related code to riscv-fusion.cc

MichielDerhaeg:


gcc/config/riscv/riscv-fusion.cc
```diff
@@ -0,0 +1,762 @@
...
+/* Implement TARGET_SCHED_MACRO_FUSION_P.  Return true if target supports
+   instruction fusion of some sort.  */
+
+bool
+riscv_macro_fusion_p (void)
+{
+  return riscv_get_fusible_ops () != RISCV_FUSE_NOTHING;
+}
+
+/* Return true iff the instruction fusion described by OP is enabled.  */
```
luismgsilva, line 46:
```suggestion
/* Return true if the instruction fusion described by OP is enabled.  */
```

luismgsilva (approved):
Looks good. Just a small typo.

MichielDerhaeg:
Merged on trunk

PR #235, opened 2026-04-10
arcv: Ignore debug instructions for fusions

exur00:
Some fusions are not detected when debug info is enabled, but are detected for the same generated code without debug info. Example: Double store fusion not detected when debug insn in between.

Sched2 output without -g:
;;	  2--> b  0: i   9 [a0]=t0
;;	  2--> b  0: i  10 [a0+0x4]=t0

Sched2 output with -g:
;;	  2--> b  0: i  12 [a0]=t0
;;	  2--> b  0: i  13 debug_marker
;;	  3--> b  0: i  14 [a0+0x4]=t0

After this patch with -g:
;;	  2--> b  0: i  12 [a0]=t0
;;	  2--> b  0: i  13 debug_marker
;;	  2--> b  0: i  14 [a0+0x4]=t0

### DIFF
--- a/gcc/config/riscv/arcv.cc
+++ b/gcc/config/riscv/arcv.cc
@@ -75,6 +75,39 @@ struct arcv_sched_state {
 
 static struct arcv_sched_state sched_state;
 
+/* Return the next possible fusible insn.  */
+
+static rtx_insn *
+arcv_next_fusible_insn (rtx_insn *insn)
+{
+  while (insn)
+    {
+      insn = NEXT_INSN (insn);
+
+      if (insn == 0)
+	break;
+
+      if (DEBUG_INSN_P (insn)
+	  || NOTE_P (insn))
+	continue;
+
+      if (NOTE_INSN_BASIC_BLOCK_P (insn))
+	return NULL;
+
+      if (GET_CODE (insn) == CODE_LABEL
+	  || GET_CODE (insn) == BARRIER
+	  || GET_CODE (PATTERN (insn)) == USE)
+	continue;
+
+      if (JUMP_TABLE_DATA_P (insn))
+	return NULL;
+
+      break;
+    }
+
+  return insn;
+}
+
 /* Return TRUE if the target microarchitecture supports macro-op
    fusion for two memory operations of mode MODE (the direction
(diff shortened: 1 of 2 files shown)


MichielDerhaeg:
Can you make some tests? Cases that were broken before but fixed now. You could use -fcompare-debug.

gcc/testsuite/gcc.target/riscv/arcv-debug-fusion-1.c
```diff
@@ -0,0 +1,13 @@
...
+/* { dg-options "-c -g -mtune=arc-v-rmx-500-series -march=rv32im -mabi=ilp32 -fdump-rtl-sched2" } */
+
+void foo(int *a) {
+    int bar = a[0] + a[1];
+    a[0] = bar;
+    a[1] = bar;
+    return;
+}
+
+/* { dg-final { scan-rtl-dump ";;\[ \t\]*(\[0-9\]+)-->\[^\n\]*\\]=.*\n(?:\[^\n\]*\n)*?;;\[ \t\]*\\1-->\[^\n\]*\\]=" "sched2" } } */
```
MichielDerhaeg: missing newline

gcc/config/riscv/arcv.cc
```diff
@@ -787,12 +787,12 @@ arcv_sched_adjust_priority (rtx_insn *insn, int priority)
   /* Bump the priority of fused load-store pairs for easier
      scheduling of the memory pipe.  The specific increase
      value is determined empirically.  */
-  if (next_insn (insn) && INSN_P (next_insn (insn))
-      && SCHED_GROUP_P (next_insn (insn))
+  rtx_insn *next = arcv_next_fusible_insn (insn);
+  if (next && INSN_P (next) && SCHED_GROUP_P (next)
```
MichielDerhaeg: Should `INSN_P` move inside arcv_next_fusible_insn?

MichielDerhaeg (changes requested):

gcc/config/riscv/arcv.cc
exur00: Done (with NONDEBUG_INSN_P)

PR #236, opened 2026-04-13
arcv: Add RHX-100 sched2 dump tests for ARC-V macro-op fusion.

luismgsilva:
... and other fixes.


luismgsilva:
I have also adapted these tests for rpx-100. But for the purpose of this upstreaming I did not include in this PR.

MichielDerhaeg:
> I have also adapted these tests for rpx-100. But for the purpose of this upstreaming I did not include in this PR.

Also because this seems to fix some things, I would apply the PR first downstream, and run some benchmarks first. This can be 1 fixup commit for marco_fusion_pair, and a separate commit for the rpx stuff.
And then we can just cherry-pick the fixup to the upstream branch.

gcc/testsuite/gcc.target/riscv/arcv-fusion-mult-add-dump.c
```diff
@@ -0,0 +1,13 @@
+/* { dg-do compile } */
+/* { dg-require-effective-target rv32 } */
+/* { dg-skip-if "" { *-*-* } { "-g" "-flto" "-O0" "-O1" "-O3" "-Oz" "-Os" } } */
+/* { dg-options "-O2 -mtune=arc-v-rhx-100-series -mno-arc-v-rmx-500-series-advanced-fusion -march=rv32im -mabi=ilp32 -fdump-rtl-sched2" } */
```
MichielDerhaeg, line 4: I have 2 issues with this.
  - The ARC_FUSE_MULT_ADD fusion shouldn't trigger with `-mno-arc-v-rmx-500-series-advanced-fusion`
  - `-mno-arc-v-rmx-500-series-advanced-fusion` doesn't exist upstream

luismgsilva: Indeed that rmx500 shouldn't be there. Forgot it during my experiments.
Either way I will also take a look on why it triggers.

luismgsilva:
> > I have also adapted these tests for rpx-100. But for the purpose of this upstreaming I did not include in this PR.
> 
> Also because this seems to fix some things, I would apply the PR first downstream, and run some benchmarks first. This can be 1 fixup commit for marco_fusion_pair, and a separate commit for the rpx stuff. And then we can just cherry-pick the fixup to the upstream branch.

Got a 74 cycles improvement on rhx and 32 cycles improvements on rpx

luismgsilva:
Closing this PR as the respective patches have been applied to `arc-2026.09-dev` in https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/237

Should be cherry-pick for `michiel/fusion-trunk-3`

PR #237, opened 2026-04-13
fixup! RISC-V: Implement riscv_macro_fusion_pair_p for Synopsys RHX-1…

luismgsilva:
…00 series.

    arcv: Fix dead memop+LUI fusion check in arcv_memop_lui_pair_p.

    arcv_memop_lui_pair_p was checking REG_P(curr) where curr is an rtx_insn*.
    REG_P tests GET_CODE(x) == REG, but an instruction RTX always has code INSN,
    JUMP_INSN, etc. — never REG.  This made is_lui unconditionally false, so the
    load+LUI and store+LUI fusion paths in arcv_macro_fusion_pair_p were never
    reached.

    Fix by checking REG_P(SET_DEST(curr_set)) instead, which correctly tests
    whether the LUI destination is a register, matching the intent of the comment:
            (set (reg:X rd) (high (const_int))).

--
    arcv: Fix ARCV_FUSE_LI_STORE mode mismatch on rv64.

    On rv64, the li destination uses DI mode while a sw store source uses SI mode.
    Since GCC uses distinct RTX objects for different modes of the same hard
    register, the pointer equality check (store_src == load_dest) always fails,
    preventing LI_STORE fusion from triggering on rv64.

    Use REGNO comparison instead of pointer equality, consistent with other register
    comparisons in arcv_macro_fusion_pair_p.

    Also remove the now-dead SUBREG_P block: it was intended to handle this mode
    narrowing case via (subreg:SI (reg:DI Rd) 0), but cleanup_subreg_operands in
    reload simplifies such subregs to plain (reg:SI Rd) before sched2 runs.  Detect
    the narrowing instead by comparing GET_MODE in the REG_P path.

### DIFF
--- a/gcc/config/riscv/arcv.cc
+++ b/gcc/config/riscv/arcv.cc
@@ -300,7 +300,7 @@ arcv_memop_lui_pair_p (rtx_insn *prev, rtx_insn *curr)
   /* Check if curr is a LUI instruction:
      - LUI via HIGH: (set (reg:X rd) (high (const_int)))
      - LUI via immediate: (set (reg:X rd) (const_int UPPER_IMM_20))  */
-  bool is_lui = (REG_P (curr)
+  bool is_lui = (REG_P (SET_DEST (curr_set))
 		&& ((get_attr_type (curr) == TYPE_MOVE
 		&& GET_CODE (SET_SRC (curr_set)) == HIGH)
 		|| (CONST_INT_P (SET_SRC (curr_set))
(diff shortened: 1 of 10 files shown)


luismgsilva:
`ARCV_FUSE_MULT_ADD` is only triggered with `-mno-arc-v-rmx-500-series-advanced-fusion`.

Looking into it.

luismgsilva:
> `ARCV_FUSE_MULT_ADD` is only triggered with `-mno-arc-v-rmx-500-series-advanced-fusion`.
> 
> Looking into it.

Fixed.

MichielDerhaeg (approved):

luismgsilva:
These patches create new failures due to `'-fcompare-debug' failure` which is fixed by this [PR](https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/235)

PR #240, opened 2026-04-21
APEX Upstreaming patch series

luismgsilva:


MichielDerhaeg (commented):
Does the dg-lto-error patch need to include some tests?
We can probably submit this asap but then we need to prove that it works by itself.

Is there a precedent for adding tests to validate the testint ifra?

luismgsilva:
> Does the dg-lto-error patch need to include some tests? We can probably submit this asap but then we need to prove that it works by itself.
> 
> Is there a precedent for adding tests to validate the testint ifra?

It makes sense. I will create test cases for this patch only.

luismgsilva:
https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/240/changes/d586196a9548942e456cf901851994c18648580a patch sent upstream: https://gcc.gnu.org/pipermail/gcc-patches/2026-May/716427.html

MichielDerhaeg:
Should the last patch not just be part of the LTO patch? The LTO patch would be incorrect without it.
Or maybe it has to be earlier in the series.

Patches are fine to send save for that.

luismgsilva:
> Should the last patch not just be part of the LTO patch? The LTO patch would be incorrect without it. Or maybe it has to be earlier in the series.
> 
> Patches are fine to send save for that.

I also thought about that. Perhaps would make more sense.

gcc/config/riscv/arcv-apex.md
```diff
@@ -0,0 +1,367 @@
...
+				 UNSPEC_ARCV_APEX_DEST_SRC0)));
+  DONE;
+})
+
+;; Used by XS, XC and XD (volatile): insn rd, src0, imm/src1
+(define_insn "riscv_arcv_apex_<APEX_DEST:mode>_ftype_<APEX_SRC0:mode>_<APEX_SRC1:mode>_v"
+  [(set (match_operand:APEX_DEST 0 "register_operand" "=r,r,r")
+	(unspec_volatile:APEX_DEST [(match_operand:SI 1 "const_int_operand" "xAVpXS,xAVpXC,xAVpXD")
+		    (match_operand:APEX_SRC0 2 "register_operand" "r,0,r")
+		    (match_operand:APEX_SRC1 3 "nonmemory_operand" "Os08,I,r")]
```
Alexehv77, line 268: try to generate some apex insn that has operand 3 a value that is large. I think this will then crash the compiler. in the expand is advised to have large predicates but in the define_insn the predicate = reunion of the constrants.

Alexehv77: after you do this then check how the compiler fixes the case where the immediate is large.

gcc/config/riscv/arcv-apex.md
```diff
@@ -0,0 +1,367 @@
...
+			reg_names[REGNO (operands[2])],
+			reg_names[REGNO (operands[3])]);
+    default:
+      gcc_unreachable ();
+  }
+}
+  [(set_attr "type" "arith,arith,arith")]
+)
+
+(define_expand "riscv_arcv_apex_dest_ftype_src0_src1_v"
```
Alexehv77, line 299: can merged with riscv_arcv_apex_dest_ftype_src0_src1

Alexehv77: also the corresponding insn

gcc/config/riscv/arcv-apex.md
```diff
@@ -0,0 +1,367 @@
...
+					       operands[3]),
+				 UNSPEC_ARCV_APEX_DEST_SRC0_SRC1_V)));
+  DONE;
+})
+
+;; Used by XS, XC and XD: insn rd, src0, imm/src1
+(define_insn "riscv_arcv_apex_<APEX_DEST:mode>_ftype_<APEX_SRC0:mode>_<APEX_SRC1:mode>"
+  [(set (match_operand:APEX_DEST 0 "register_operand" "=r,r,r")
+	(unspec:APEX_DEST [(match_operand:SI 1 "const_int_operand" "xAVpXS,xAVpXC,xAVpXD")
+		    (match_operand:APEX_SRC0 2 "register_operand" "r,0,r")
```
Alexehv77, line 320: Here (and where else you use input operand 0) is a problem.
after reload_completed there may not be possible to have operand 0 same as operand 0 of your instruction - imagine the case where the input operand 2 is used later by another instruction. But at the same time you overwrite it as your operand 0 is output.
You need to make a split and in that split next to your instruction to add a move from reg 2 to reg 0. as this move is a set will be later optimized away if not needed.
You can look to my push for target arc64 where i have several define_insn_and_split

PR #241, opened 2026-04-23
Riscv hoist address ladder

Alexehv77:


gcc/config/riscv/riscv-matrix-opt.cc
```diff
@@ -0,0 +1,346 @@
+/* This pass identifies addressing patterns in memory-access intensive 
+ * kernels. In such kernels an index is frequently scaled and added to a 
+ * base pointer to access memory followed by an immediate index update.
+ * The pass hoists the pointer arithmetic out of the loop header converting
+ * index addressing into induction variables. */
```
MichielDerhaeg, line 5: Isn't this what ivopts does? What is the point of doing it this way?

MichielDerhaeg:
Replaced by https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/261

PR #243, opened 2026-04-28
Rebase downstream fusion patches to upstream

luismgsilva:
This PR rebases the following downstream patches into fixups to the current fusion trunk cleanup branch.
- 2d1beeba18f9e5a0c254a4d08307691107255e9b
- 4c75c463724bd583e49c54fd1b1531d42edd1051
- 83600fe0bc7e410596a79bfb1fa2effc76f30c82
- 0db4e8b8f26607b18fbb7d9c2abe27d1f83d3990

A RMX-500 test case from 2d1beeba18f9e5a0c254a4d08307691107255e9b is not included in this PR.

### DIFF
--- a/gcc/config/riscv/arcv.cc
+++ b/gcc/config/riscv/arcv.cc
@@ -665,7 +665,7 @@ arcv_sched_reorder2 (rtx_insn **ready, int *n_readyp)
 	     && get_attr_type (ready[*n_readyp - i]) != TYPE_STORE
 	     && !SCHED_GROUP_P (ready[*n_readyp - i])
 	     && (!next_insn || !SCHED_GROUP_P (next_insn)))
-	    || (next_insn && NONDEBUG_INSN_P (next_insn)
+	    || (next_insn
 		&& recog_memoized (next_insn) >= 0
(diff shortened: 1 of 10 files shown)


luismgsilva:
These patches has been applied to our downstream `arc-2026.09-dev` branch.

EDIT: ~I will convert them all to fixups.~ Done

gcc/config/riscv/arcv.cc
```diff
@@ -75,6 +75,38 @@ struct arcv_sched_state {
 
 static struct arcv_sched_state sched_state;
 
+/* Return the next possible fusible insn.  */
+
+static rtx_insn *
+arcv_next_fusible_insn (rtx_insn *insn)
```
MichielDerhaeg, line 81: Discussed this new implementation a bit with @exur00 
It is likely quite wrong and can be simplified a bit.
I suspect it is functionally correct though, but we will have to fix it before submitting again.

MichielDerhaeg: wrong as in, lots of redundant and dead code

MichielDerhaeg (approved):
Fine to merge, but next_fusible_insn will need to be fixed later

PR #244, opened 2026-04-30
arcv: add mul-mem bypasses

exur00:


gcc/config/riscv/arcv-rhx100.md
```diff
@@ -104,3 +104,6 @@
 
 (define_bypass 3 "arcv_rhx100_mpy32_insn" "arcv_rhx100_mpy*_insn")
 (define_bypass 3 "arcv_rhx100_mpy32_insn" "arcv_rhx100_div_insn")
+(define_bypass 1 "arcv_rhx100_mpy32_insn" "arcv_rhx100_store_insn" "riscv_store_data_bypass_p")
+(define_bypass 7 "arcv_rhx100_mpy32_insn" "arcv_rhx100_store_insn")
```
MichielDerhaeg, line 108: Doesn't this have to check the inverse of the predicate? now they overlap.

exur00: That is actually fine
> If there are more one bypass with the same output and input insns, the chosen bypass is the first bypass with a guard in description whose guard function returns nonzero. If there is no such bypass, then bypass without the guard function is chosen. ( https://gcc.gnu.org/onlinedocs/gccint/Processor-pipeline-description.html )

MichielDerhaeg (approved):

PR #247, opened 2026-05-06
arcv: Fix ls_update for float loads/stores

MichielDerhaeg:


luismgsilva (approved):
Looks good. I assume this is to be also applied into `michiel/fusion-trunk-3`

MichielDerhaeg:
> Looks good. I assume this is to be also applied into `michiel/fusion-trunk-3`

Let's wait with the float stuff. I would still evaluate this with some benchmarks. We can send a new set of patches upstream with float support.

PR #248, opened 2026-05-06
arcv: Add floats for double load/store fusion

MichielDerhaeg:


MichielDerhaeg:
Superseded by https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/192

PR #249, opened 2026-05-07
RISC-V: Don't emit cm.popret when zcmp + zicfiss is enabled

MichielDerhaeg:


MichielDerhaeg:
Cherry-picked from upstream: 83893a498ed39de5f46a872a94892fa3507a9148

PR #250, opened 2026-05-08
arcv: simplify arcv_next_fusible_insn

exur00:
The removed checks were impossible to trigger, !NONDEBUG_INSN_P checked that already. The early are moved forward because they were unreachable.


MichielDerhaeg (approved):
Ok if ci passes

PR #251, opened 2026-05-08
arcv: FPU latencies for RHX-100

MichielDerhaeg:


exur00:
FMADD seems like an instruction that could be used often, maybe it's worth creating a guard to also model the FMADD/DMADD 3rd operand bypass?

exur00 (approved):

MichielDerhaeg:
> FMADD seems like an instruction that could be used often, maybe it's worth creating a guard to also model the FMADD/DMADD 3rd operand bypass?

Definitely. Perhaps in a subsequent patch though.

luismgsilva (approved):

PR #252, opened 2026-05-11
fix for P10019563-77915

Alexehv77:
fix to ensure the that the hs5x target produces correct code despite lacking 64-bit load/store instructions.
I made use of the split mechanism enabled by the # keyword which forces the compiler to look for a define_split. Then the split pattern that takes a single 64bit memory move and breaks it into two 32-bit moves using gen_lowpart and gen_highpart.
I then tried to run dejagnu and found failures that were not due to my patch.

### DIFF
--- a/gcc/config/arc64/arc64.cc
+++ b/gcc/config/arc64/arc64.cc
@@ -512,7 +512,18 @@ arc64_save_reg_p (int regno)
   return false;
 }
 
-/* Compute the frame info.  */
+/* This function calculates the stack frame before the compiler outputs
+   the prologue and epilogue instructions.
+   It takes care of the stack allocation calculating frame sizes and
+   tracking memory byte offsets for every callee saved register.
+
+   On 32 bit like hs5x running double precision FPU logic without a native
+   64bit wide load/store unit (!TARGET_LL64) tje 64 bit
+   double words (DFmode) cannot be transferred atomically.  They must be split.
+   This function dynamically assigns a separate, individual 32 bit word slot
+   (UNITS_PER_WORD) to both halves of an FPU register pair sequentially.
+   By manually stepping the iterator (regno += 2) we take care that the low
+   half and high half are allocated next to eachother.  */
 
 static void
 arc64_compute_frame_info (void)
@@ -527,23 +538,41 @@ arc64_compute_frame_info (void)
 
   if (!ARC_NAKED_P(cfun->machine->fn_type))
     {
+      /* Initialize the offset tracking array to unallocated (-1) status.  */
+      for (regno = 0; regno < FIRST_PSEUDO_REGISTER; regno++)
+	frame->reg_offset[regno] = -1;
+
       /* Find out which GPR need to be saved.  */
-      for (regno = R0_REGNUM, offset = 0;
-	   regno <= F31_REGNUM;
-	   regno++)
+      /* Move the iterator step inside the loop body to control
+	 64bit FPU pairs dynamically when compiling for hs5x.  */
+      for (regno = R0_REGNUM, offset = 0; regno <= F31_REGNUM;)
 	if (arc64_save_reg_p (regno))
 	  {
-	    /* TBI: probably I need to make the saving of the FP registers
-	       separate bulk from GPIs such that I can use latter on enter/leave
-	       instruction seamlessly (i.e. first save FPregs/latter GPI, the
-	       leave return feature will not work).  */
-	    /* TBI: the FPUS only configuration is having only 32bit registers,
-	       thus I can stack 2 FP registers in one stack slot ;).  */
+	    /* Separate FPU registers when double prec is active and the
+	       hardware lacks native 64 bit load/store support
+	       (!TARGET_LL64). */
+	    if (ARC64_HAS_FP_BASE && FP_REGNUM_P (regno)
+		&& ARC64_HAS_FPUD && !TARGET_LL64)
+	      {
+		frame->reg_offset[regno] = offset;
+		offset += UNITS_PER_WORD;
+
+		frame->reg_offset[regno + 1] = offset;
+		offset += UNITS_PER_WORD;
+
+		regno += 2;
+		continue;
+	      }
+
 	    frame->reg_offset[regno] = offset;
 	    offset += UNITS_PER_WORD;
+	    regno++;
 	  }
 	else
-	  frame->reg_offset[regno] = -1;
+	  {
+	    frame->reg_offset[regno] = -1;
+	    regno++;
+	  }
 
       /* Check if we need to save the return address.  */
       if (!crtl->is_leaf
@@ -610,16 +639,41 @@ frame_stack_add (HOST_WIDE_INT offset)
 					    offset)));
 }
 
-/* Helper for prologue: emit frame store with pre_modify or pre_dec to
-   save register REG on stack.  An initial offset OFFSET can be passed
-   to the function.  If a DISPLACEMENT is defined, it will be used to
-   generate pre_modify instead of pre_dec.  */
-
 static HOST_WIDE_INT
 frame_save_reg (rtx reg, HOST_WIDE_INT offset, HOST_WIDE_INT displacement)
 {
   rtx addr, tmp;
 
+  if (!TARGET_LL64 && REG_P (reg) && FP_REGNUM_P (REGNO (reg)))
+    {
+      machine_mode mode = GET_MODE (reg);
+      HOST_WIDE_INT mode_size = GET_MODE_SIZE (mode);
+
+      HOST_WIDE_INT actual_offset = offset - 
+	      (displacement ? displacement : mode_size);
+
+      rtx flat_addr = plus_constant (Pmode, stack_pointer_rtx, actual_offset);
+      addr = gen_frame_mem (GET_MODE (reg), flat_addr);
+
+      /* stack_pointer_rtx will determined the mode of the add - depending if 
+       * the core is 64/32 bit. */
+      rtx sp_adjust = emit_insn (gen_add3_insn (stack_pointer_rtx,
+                                                stack_pointer_rtx, 
+                                                GEN_INT (actual_offset)));
+      
+      RTX_FRAME_RELATED_P (sp_adjust) = 1;
+      add_reg_note (sp_adjust, REG_CFA_ADJUST_CFA, gen_rtx_SET (
+        stack_pointer_rtx, plus_constant (Pmode,
+                stack_pointer_rtx, actual_offset)));
+
+      tmp = emit_move_insn (addr, reg);
+      RTX_FRAME_RELATED_P (tmp) = 1;
+
+      return (displacement ? displacement : GET_MODE_SIZE (GET_MODE (reg)))
+              - offset;
+    }
+
+  /* Original code for hs6x target registers.  */
   if (offset)
     {
       tmp = plus_constant (Pmode, stack_pointer_rtx,
@@ -659,21 +713,64 @@ arc64_save_callee_saves (void)
   HOST_WIDE_INT frame_allocated = 0;
   rtx reg;
 
-  for (regno = F31_REGNUM; regno >= R0_REGNUM; regno--)
+  for (regno = F31_REGNUM; regno >= R0_REGNUM;)
     {
       HOST_WIDE_INT disp = 0;
       if (frame->reg_offset[regno] == -1
 	  /* Hard frame pointer is saved in a different place.  */
 	  || (frame_pointer_needed && regno == R27_REGNUM)
-	  /* blink register is saved in a different place.  */
 	  || (regno == BLINK_REGNUM))
-	continue;
+	{
+	  regno--;
+	  continue;
+	}
 
       save_mode = word_mode;
       if (ARC64_HAS_FP_BASE && FP_REGNUM_P (regno))
 	{
 	  save_mode = ARC64_HAS_FPUD ? DFmode : SFmode;
 	  disp = UNITS_PER_WORD;
+
+	  if (save_mode == DFmode && !TARGET_LL64)
+	    {
+	      /* Force pass through general purpose registers.
+		 Move FPU data to r12:r13, allocate stack space, and
+		 store via core GPR paths.  */
+	      int even_base = (regno % 2 == 0) ? regno : regno - 1;
+	      rtx sp_reg = stack_pointer_rtx;
+
+	      rtx gpr_lo = gen_rtx_REG (SImode, R12_REGNUM); /* r12.  */
+	      rtx gpr_hi = gen_rtx_REG (SImode, R13_REGNUM); /* r13.  */
+	      rtx fpu_lo = gen_rtx_REG (SImode, even_base);
+	      rtx fpu_hi = gen_rtx_REG (SImode, even_base + 1);
+
+	      emit_move_insn (gpr_lo, fpu_lo);
+	      emit_move_insn (gpr_hi, fpu_hi);
+
+	      rtx insn_sp = emit_insn (gen_addsi3 (sp_reg, sp_reg,
+				      GEN_INT (-8)));
+	      RTX_FRAME_RELATED_P (insn_sp) = 1;
+	      add_reg_note (insn_sp, REG_CFA_ADJUST_CFA, gen_rtx_SET (sp_reg,
+				      plus_constant (SImode, sp_reg, -8)));
+
+	      rtx insn_lo = emit_move_insn (gen_frame_mem
+			      (SImode, sp_reg), gpr_lo);
+	      RTX_FRAME_RELATED_P (insn_lo) = 1;
+
+	      rtx insn_hi = emit_move_insn (gen_frame_mem (SImode,
+			    plus_constant (SImode, sp_reg, 4)), gpr_hi);
+	      RTX_FRAME_RELATED_P (insn_hi) = 1;
+
+	      add_reg_note (insn_lo, REG_CFA_OFFSET, gen_rtx_SET (
+			    gen_frame_mem (SImode, sp_reg), fpu_lo));
+	      add_reg_note (insn_hi, REG_CFA_OFFSET, gen_rtx_SET (
+	      gen_frame_mem (SImode, plus_constant (SImode, sp_reg, 4)),
+	      fpu_hi));
+
+	      frame_allocated += 8;
+	      regno = even_base - 1;
+	      continue;
+	    }
 	}
       else if (regno >= 1
 	       && (((regno - 1) % 2) == 0)
@@ -696,6 +793,7 @@ arc64_save_callee_saves (void)
       reg = gen_rtx_REG (save_mode, regno);
       frame_allocated += frame_save_reg (reg, offset, disp);
       offset = 0;
+      --regno;
     }
 
   /* Save BLINK if required.  */
@@ -723,15 +821,44 @@ arc64_save_callee_saves (void)
   return frame_allocated;
 }
 
-/* Helper for epilogue: emit frame load with post_modify or post_inc
-   to restore register REG from stack.  The initial offset is passed
-   via OFFSET.  */
+/* This function emits the individual movement instructions
+   necessary during the function epilog to restore a single callee saved
+   register from its slot in the stack frame.  It also handles the tracking
+   of stack pointer deallocation.  It returns the number of bytes out off
+   the stack.  */
 
 static HOST_WIDE_INT
 frame_restore_reg (rtx reg, HOST_WIDE_INT displacement)
 {
   rtx addr, insn, tmp;
 
+  /* Try to force flat linear pop sequences for FPU registers.  */
+  if (!TARGET_LL64 && REG_P (reg) && FP_REGNUM_P (REGNO (reg)))
+    {
+      HOST_WIDE_INT actual_disp = displacement ? displacement :
+				  GET_MODE_SIZE (GET_MODE (reg));
+
+      /* Load from the current flat frame stack slice address without
+	 post increment.  */
+      rtx flat_addr = plus_constant (Pmode, stack_pointer_rtx, displacement);
+      addr = gen_frame_mem (GET_MODE (reg), flat_addr);
+
+      insn = emit_move_insn (reg, addr);
+      RTX_FRAME_RELATED_P (insn) = 1;
+      add_reg_note (insn, REG_CFA_RESTORE, reg);
+
+      rtx sp_adjust = emit_insn (gen_addsi3 (stack_pointer_rtx,
+			  stack_pointer_rtx, GEN_INT (actual_disp)));
+      RTX_FRAME_RELATED_P (sp_adjust) = 1;
+      add_reg_note (sp_adjust, REG_CFA_ADJUST_CFA,
+		    gen_rtx_SET (stack_pointer_rtx, plus_constant (Pmode,
+				    stack_pointer_rtx, actual_disp)));
+
+      return actual_disp;
+    }
+
+  /* Original native logic continues untouched for integer and
+     hs6x target registers.  */
   if (displacement)
     {
       tmp = plus_constant (Pmode, stack_pointer_rtx, displacement);
@@ -759,8 +886,6 @@ frame_restore_reg (rtx reg, HOST_WIDE_INT displacement)
 
   return displacement ? displacement : GET_MODE_SIZE (GET_MODE (reg));
 }
-
-/* ARC' epilogue restore regs routine.  */
 
 static HOST_WIDE_INT
 arc64_restore_callee_saves (bool sibcall_p ATTRIBUTE_UNUSED)
@@ -796,7 +921,7 @@ arc64_restore_callee_saves (bool sibcall_p ATTRIBUTE_UNUSED)
       frame_deallocated += frame_restore_reg (reg, 0);
     }
 
-  for (regno = R0_REGNUM; regno <= F31_REGNUM; regno++)
+  for (regno = R0_REGNUM; regno <= F31_REGNUM;)
     {
       HOST_WIDE_INT disp = 0;
       bool double_load_p = false;
@@ -806,13 +931,55 @@ arc64_restore_callee_saves (bool sibcall_p ATTRIBUTE_UNUSED)
 	  || (frame_pointer_needed && regno == R27_REGNUM)
 	  /* blink register has been restored.  */
 	  || (regno == BLINK_REGNUM))
-	continue;
+	{
+	  regno++;
+	  continue;
+	}
 
       restore_mode = word_mode;
       if (ARC64_HAS_FP_BASE && FP_REGNUM_P (regno))
 	{
 	  restore_mode = ARC64_HAS_FPUD ? DFmode : SFmode;
 	  disp = UNITS_PER_WORD;
+
+	  if (restore_mode == DFmode && !TARGET_LL64)
+	    {
+	      int even_base = (regno % 2 == 0) ? regno : regno - 1;
+	      rtx sp_reg = stack_pointer_rtx;
+
+	      rtx gpr_lo = gen_rtx_REG (SImode, R12_REGNUM); /* r12.  */
+	      rtx gpr_hi = gen_rtx_REG (SImode, R13_REGNUM); /* r13.  */
+	      rtx fpu_lo = gen_rtx_REG (SImode, even_base);
+	      rtx fpu_hi = gen_rtx_REG (SImode, even_base + 1);
+
+	      /* extract data back into core pipeline registers first via
+		 standard loads (ld) */
+	      rtx insn_lo = emit_move_insn (gpr_lo,
+			      gen_frame_mem (SImode, sp_reg));
+	      RTX_FRAME_RELATED_P (insn_lo) = 1;
+	      add_reg_note (insn_lo, REG_CFA_RESTORE, fpu_lo);
+
+	      rtx insn_hi = emit_move_insn (gpr_hi, gen_frame_mem (SImode,
+				      plus_constant (SImode, sp_reg, 4)));
+	      RTX_FRAME_RELATED_P (insn_hi) = 1;
+	      add_reg_note (insn_hi, REG_CFA_RESTORE, fpu_hi);
+
+	      /* Move data from GPRs back to the final FPU register halves.  */
+	      emit_move_insn (fpu_lo, gpr_lo);
+	      emit_move_insn (fpu_hi, gpr_hi);
+
+	      /* get stack allocations manually.  */
+	      rtx insn_sp = emit_insn (gen_addsi3 (sp_reg, sp_reg,
+				      GEN_INT (8)));
+	      RTX_FRAME_RELATED_P (insn_sp) = 1;
+	      add_reg_note (insn_sp, REG_CFA_ADJUST_CFA,
+			    gen_rtx_SET (sp_reg, plus_constant (
+					    SImode, sp_reg, 8)));
+
+	      frame_deallocated += 8;
+	      regno = even_base + 2;
+	      continue;
+	    }
 	}
       else if ((regno % 2) == 0
 	       && (!frame_pointer_needed || ((regno + 1) != R27_REGNUM))
@@ -837,7 +1004,9 @@ arc64_restore_callee_saves (bool sibcall_p ATTRIBUTE_UNUSED)
       frame_deallocated += frame_restore_reg (reg, disp);
 
       if (double_load_p)
-	regno++;
+	 regno += 2;
+      else
+	 regno++;
     }
 
   return frame_deallocated;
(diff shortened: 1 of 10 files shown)


gcc/config/arc64/arc64.md
```diff
@@ -927,24 +930,79 @@ xorl"
...
+;;   fst<sizef>%U0\\t%1,%0
+;;   fmv<fmvftab>2<fmvitab>\\t%0,%1
+;;   fmv<fmvitab>2<fmvftab>\\t%0,%1
+;;   mov<mcctab>\\t%0,%1
+;;   mov<mcctab>\\t%0,%1
+;;   ld<slfp>%U1\\t%0,%1
+;;   st<slfp>%U0\\t%1,%0"
+;;  [(set_attr "type" "fmov,ld,st,move,move,move,move,ld,st")
+;;   (set_attr "length" "4,*,*,4,4,4,8,*,*")])
+
```
MichielDerhaeg: Commented out?

gcc/config/arc64/arc64.md
```diff
@@ -968,12 +1026,127 @@ xorl"
...
+;;   #
+;;   #"
+;;  [(set_attr "type" "move,ld,st") 
+;;   (set (attr "enabled") 
+;;     (cond [(eq_attr "type" "move" ) 
+;;            (const_string "yes")
+;;            (match_test "TARGET_LL64")    
+;;            (const_string "yes") ]
+;;           (const_string "no")))
+;;   (set_attr "length" "4,8,8")])
```
MichielDerhaeg: Idem.

gcc/testsuite/gcc.target/arc64/pr10019563-77915.c
```diff
@@ -0,0 +1,13 @@
...
+
+int test0(){
+    float a, b;
+    double c, d;    
+    a = (float)rand()/(float)(RAND_MAX);
+    b = (float)rand()/(float)(RAND_MAX);    
+    c = (double)rand()/(double)(RAND_MAX);
+    d = (double)rand()/(double)(RAND_MAX);    
+    return (int)((double)a + (double)b + c + d);
+}
```
MichielDerhaeg, line 14: Should this be a torture test? You're not checking the result, nor the generated code.

gcc/testsuite/gcc.target/arc64/pr10019563-77915_m.c
```diff
@@ -0,0 +1,12 @@
...
+/* Standard success is return 0 */
+volatile double a = 10.0;
+volatile double b = 5.0;
+
+int main() {
+    double res = a - b;
+    if (res == 5.0)
+        return 0;
+    return 1;
+}
```
MichielDerhaeg, line 16: This should be in gcc.target/arc64. this will fail when build for other targets.

gcc/testsuite/gcc.target/arc64/pr10019563-77915_m.c
```diff
@@ -0,0 +1,12 @@
...
+/* { dg-options "-O2 -mcpu=hs5x -mfpu=fpud" } */
+/* Standard success is return 0 */
+volatile double a = 10.0;
+volatile double b = 5.0;
+
+int main() {
+    double res = a - b;
+    if (res == 5.0)
+        return 0;
+    return 1;
```
MichielDerhaeg, line 15: I would also check produced instructions. This test succeeds if it does floating-point emulation.

gcc/config/arc64/arc64.md
```diff
@@ -968,12 +1026,127 @@ xorl"
...
+          hi_dest = r1*/
+          emit_move_insn (lo_dest, lo_src);
+          emit_move_insn (hi_dest, hi_src);
+      }
+
+    /* emit post increments  */
+    if (MEM_P (src) && GET_CODE (XEXP (src, 0)) == POST_INC)
+      emit_insn (gen_adddi3 (XEXP (XEXP (src, 0), 0), XEXP (XEXP (src, 0), 0), GEN_INT (8)));
+    if (MEM_P (dest) && GET_CODE (XEXP (dest, 0)) == POST_INC)
+      emit_insn (gen_adddi3 (XEXP (XEXP (dest, 0), 0), XEXP (XEXP (dest, 0), 0), GEN_INT (8)));
```
MichielDerhaeg: Lot's of code, perhaps better to move it to arc64.cc and call it here.

gcc/testsuite/gdc.dg/torture/pr102139.c
```diff
@@ -0,0 +1,43 @@
+/* { dg-do run } */
```
MichielDerhaeg: Where does this test come from?

gcc/config/arc64/arc64.cc
```diff
@@ -674,6 +673,26 @@ arc64_save_callee_saves (void)
...
+	    {
+	      /* Save the lower 4-byte sub-register piece (regno) */
+	      rtx reg_lo = gen_rtx_REG (SFmode, regno);
+	      frame_allocated += frame_save_reg (reg_lo, offset, disp);
+	      offset = 0;
+
+	      /* Save the upper 4-byte sub-register piece (regno + 1).
+		 The displacement or offset is updated implicitly by 
+		 the frame tracker inside frame_save_reg. */
+	      rtx reg_hi = gen_rtx_REG (SFmode, regno + 1);
```
MichielDerhaeg: First, shouldn't this be regno - 1?
Also shouldn't you additionally increment/decrement regno, because now you're handling a register twice I think.

gcc/config/arc64/arc64.md
```diff
@@ -924,27 +927,56 @@ xorl"
...
+       */
+      case 8:
+      /* option 8 (*Ustor, *r) - is a Store. 
+         It moves 64 bits from a GPR pair back into Memory.
+       */
+        if (GET_MODE_SIZE (<MODE>mode) == 8 && !TARGET_LL64)
+          /*for hx5s with fp unit that supports double precision fp
+            we trigger the splitter*/
+          return "#";
+        return (which_alternative == 7) ?
```
MichielDerhaeg: Can we clean this up a bit, and get it closer to the original by making this a define_insn_and_split with this as the split condition?

gcc/config/arc64/arc64.md
```diff
@@ -999,6 +1133,40 @@ xorl"
     (set_attr "length" "2,2,8,8,4,4,8,8,8,8,2,*,2,*")]
 )
 
+(define_insn "*arc64_movdi"
```
MichielDerhaeg: 2 definitions of `*arc64_movdi` is problematic I think.

gcc/testsuite/gcc.target/arc64/pr10019563-77915.c
MichielDerhaeg: On second thought. Does this work? It can't run if there is no `main`?

Alexehv77: the main was below. i moved those tests to the arc64 directory. the test is the original test that Luis added to the jira ticket. This test was exposing a compilation error so i decided to let it as was. It also has the name of the jira ticket.

gcc/testsuite/gcc.target/arc64/pr10019563-77915_m.c
Alexehv77: i added the test pr10019563-77915_align.c that checks generated code.

Alexehv77 (commented):
I updated the tests and cleaned the remaining code

gcc/config/arc64/arc64.cc
```diff
@@ -574,7 +590,7 @@ arc64_compute_frame_info (void)
      FIXME! FPR registers.  */
   frame->saved_regs_size = ARC64_STACK_ALIGN (offset);
 
-  /* 4. Size of the callee-allocated area for pretend stack
+  /* 4. Size of the callee-allocated area for pretend stack
```
luismgsilva: Trailing space?

gcc/config/arc64/arc64.cc
```diff
@@ -647,8 +686,65 @@ frame_save_reg (rtx reg, HOST_WIDE_INT offset, HOST_WIDE_INT displacement)
...
 }
 
-/* ARC prologue saving regs routine.   */
+/*add updated comment*/
+static HOST_WIDE_INT
+frame_restore_reg (rtx reg, HOST_WIDE_INT displacement)
+{
+  rtx addr, insn, tmp;
+
+  /* try to force flat linear pop sequences for FPU registers */
```
luismgsilva: Comments not in GNU style.
```suggestion
  /* Try to force flat linear pop sequences for FPU registers.  */
```
Apply this for the rest of the comments.

gcc/config/arc64/arc64.cc
```diff
@@ -512,8 +512,6 @@ arc64_save_reg_p (int regno)
   return false;
 }
 
-/* Compute the frame info.  */
```
luismgsilva, line 515: Why did you remove the comment?

gcc/config/arc64/arc64.cc
```diff
@@ -610,20 +626,43 @@ frame_stack_add (HOST_WIDE_INT offset)
 					    offset)));
 }
 
-/* Helper for prologue: emit frame store with pre_modify or pre_dec to
```
luismgsilva, line 613: Idem.

gcc/config/arc64/arc64.cc
```diff
@@ -723,45 +854,6 @@ arc64_save_callee_saves (void)
...
-  else
-    add_reg_note (insn, REG_CFA_ADJUST_CFA,
-		  gen_rtx_SET (stack_pointer_rtx,
-			       plus_constant (Pmode, stack_pointer_rtx,
-					      GET_MODE_SIZE (GET_MODE (reg)))));
-
-  return displacement ? displacement : GET_MODE_SIZE (GET_MODE (reg));
-}
-
-/* ARC' epilogue restore regs routine.  */
```
luismgsilva, line 763: Idem.

gcc/config/arc64/arc64.cc
```diff
@@ -806,26 +895,62 @@ arc64_restore_callee_saves (bool sibcall_p ATTRIBUTE_UNUSED)
...
 	{
 	  restore_mode = ARC64_HAS_FPUD ? DFmode : SFmode;
 	  disp = UNITS_PER_WORD;
+
+	  if (restore_mode == DFmode && !TARGET_LL64)
+	    {
+	      int even_base = (regno % 2 == 0) ? regno : regno - 1;
+	      rtx sp_reg = stack_pointer_rtx;
+
+	      rtx gpr_lo = gen_rtx_REG (SImode, 2); /* r2 */
```
luismgsilva: Is it safe to hardcode register numbers? If r2/r3 are live at these points this will silently clobber them.

gcc/config/arc64/arc64.md
```diff
@@ -865,34 +870,207 @@ xorl"
...
+   st\\t0,%0
+   st%U0\\t0,%0"
+  [(set_attr "type" "st")
+   (set_attr "length" "4,4")])
+
+(define_insn "*subsi3_insn"
+  [(set (match_operand:SI 0 "register_operand"           "=r,r,r,r")
+        (minus:SI (match_operand:SI 1 "register_operand"  "0,r,0,r")
+                  (match_operand:SI 2 "nonmemory_operand" "r,r,i,i")))]
+  "TARGET_64BIT || !TARGET_64BIT"
```
luismgsilva: This is always true.

gcc/config/arc64/arc64.cc
```diff
@@ -6721,6 +6874,72 @@ arc64_expand_vector_init (rtx target, rtx vals)
     }
 }
 
+static int
+arc64_contains_symbol_ref_callback (rtx *x, void *data ATTRIBUTE_UNUSED)
```
luismgsilva: I dont see this function being called anywhere.

gcc/config/arc64/arc64.cc
```diff
@@ -610,20 +626,43 @@ frame_stack_add (HOST_WIDE_INT offset)
...
+
+      return (displacement ? displacement : GET_MODE_SIZE (GET_MODE (reg))) - offset;
+    }
+
+  /* Original code for hs6x target registers */
   if (offset)
     {
       tmp = plus_constant (Pmode, stack_pointer_rtx,
-			       offset - GET_MODE_SIZE (GET_MODE (reg)));
+			   offset - GET_MODE_SIZE (GET_MODE (reg)));
```
luismgsilva: Why did you touch this?

gcc/config/arc64/arc64.cc
```diff
@@ -610,20 +626,43 @@ frame_stack_add (HOST_WIDE_INT offset)
...
 frame_save_reg (rtx reg, HOST_WIDE_INT offset, HOST_WIDE_INT displacement)
 {
   rtx addr, tmp;
 
+  if (!TARGET_LL64 && REG_P (reg) && FP_REGNUM_P (REGNO (reg)))
+    {
+      HOST_WIDE_INT actual_offset = offset;
+      
+      if (displacement)
+	actual_offset = -displacement;
```
luismgsilva: Shouldnt this be ?
```suggestion
	actual_offset = offset - displacement;
```

When `offset != 0`, the current code ignores it in the SP adjustment but still subtracts it in the return value (See line [658](https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/252/changes#diff-dfd12f56fd5a93b01461dca7c6ec4cd431f40c7b9abf5db20ed682fa553b16eaR658)). The original path (See line [665)](https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/252/changes#diff-dfd12f56fd5a93b01461dca7c6ec4cd431f40c7b9abf5db20ed682fa553b16eaR665) checks offset first and uses `offset - GET_MODE_SIZE(reg)`, keeping SP movement and the return value in sync.

gcc/config/arc64/arc64.cc
```diff
@@ -806,26 +895,62 @@ arc64_restore_callee_saves (bool sibcall_p ATTRIBUTE_UNUSED)
...
+	      continue;
+	    }
 	}
       else if ((regno % 2) == 0
 	       && (!frame_pointer_needed || ((regno + 1) != R27_REGNUM))
 	       && (frame->reg_offset[regno + 1] != -1
 		   && ((regno + 1) != BLINK_REGNUM)))
 	{
-	  /* Use 64-bit double loads for context restoring.  */
+	  /* Make use of 64 bit double loads for context restoring  */
```
luismgsilva: is there a need to change this comment?

gcc/config/arc64/arc64.cc
```diff
@@ -806,26 +895,62 @@ arc64_restore_callee_saves (bool sibcall_p ATTRIBUTE_UNUSED)
...
 	  if (!TARGET_64BIT && TARGET_LL64)
-	    {
+	    { 
 	      restore_mode = DImode;
 	      double_load_p = true;
 	    }
-	  /* Use 128-bit double loads for context restoring.  */
```
luismgsilva, line 828: Why was this comment removed?

gcc/config/arc64/arc64.cc
```diff
@@ -837,12 +962,16 @@ arc64_restore_callee_saves (bool sibcall_p ATTRIBUTE_UNUSED)
...
       if (double_load_p)
+	regno += 2;
+      else
 	regno++;
     }
 
   return frame_deallocated;
 }
 
+
```
luismgsilva: Extra spaces?

gcc/config/arc64/arc64.cc
```diff
@@ -6135,8 +6289,7 @@ arc64_split_double_move (rtx *operands, machine_mode mode)
   machine_mode mvmode = smallest_int_mode_for_size (BITS_PER_WORD).require ();
 
   /* Maximum size handled is twice UNITS_PER_WORD.  */
-  gcc_assert (iregs <= 2);
-
+    gcc_assert(iregs <= 2);
```
luismgsilva: Why did you touch this? You removed a space and a newline

luismgsilva (changes requested):
Please check the GNU style on the patch: https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/actions/runs/26152174134/job/76922124759?pr=252

Alexehv77 (commented):
i changed those files please have a look

gcc/config/arc64/arc64.md
Alexehv77: i removed the not needed one

gcc/config/arc64/arc64.md
Alexehv77: i removed it

gcc/config/arc64/arc64.md
Alexehv77: removed it

gcc/testsuite/gcc.target/arc64/pr10019563-77915_m.c
Alexehv77: placed it in arc64

gcc/testsuite/gdc.dg/torture/pr102139.c
Alexehv77: is fine now. i just checked it

gcc/config/arc64/arc64.cc
Alexehv77: this code is gone

gcc/config/arc64/arc64.md
Alexehv77: can not merge this define_insn with the splitter as the splitter applies only for some of the define_insn clauses.

gcc/config/arc64/arc64.cc
Alexehv77: i took care of it

gcc/config/arc64/arc64.cc
Alexehv77: this is inside the restore_callee_saves - which is the function that (as its name says it) deals with restoring the callee saved registers. Those are the registers that were needed to compile the callee next to the caller saved which are offered to the callee by the caller.
Thus I can use any of the caller saved registers excepting those used for passing arguments. I thought r0 and r1 are used for in/out arguments. For now i use 12 and 13 which according to the abi are safe to be used here. Otherwise I will implement some late stack spilling save/restore directly inside the epilogue code generation block. I can use for example the r0/r1 to save them, then use as scratch for FP and then restore them. Those are core cases that need to be checked and fro which tests need to be made...

gcc/config/arc64/arc64.md
Alexehv77: i removed the conditions

gcc/config/arc64/arc64.cc
Alexehv77: is left over from the next function insn_has_symbol. this i added to provent the delay slot scheduler to get long immediates inside the delay slots.

gcc/config/arc64/arc64.cc
Alexehv77: no idea. i reverted it

gcc/config/arc64/arc64.cc
Alexehv77: i am busy checking this.

gcc/config/arc64/arc64.cc
Alexehv77: i think i changed it during the gnu style checking.

gcc/config/arc64/arc64.cc
Alexehv77: added it

gcc/config/arc64/arc64.cc
Alexehv77: reverted it

gcc/config/arc64/arc64.cc
Alexehv77: the comment compute the frame info for a function called compute_frame_info is superfluous. I made changes in this function and my intend was to add extra comments which i did now.

gcc/config/arc64/arc64.cc
Alexehv77: as i modified it i added a comment that i have to add a comment. i did add it meanwhile.

gcc/config/arc64/arc64.cc
Alexehv77: i wanted to rewrite the comment. for now i kept the original

gcc/config/arc64/arc64.cc
```diff
@@ -946,18 +1118,19 @@ arc64_legitimate_address_1_p (machine_mode mode,
 	 memory accesses are splited anyhow.  */
       if (GET_MODE_SIZE (mode) == UNITS_PER_WORD * 2)
 	{
-	  /* 32-bit and no double loads?  */
+	  /* 32-bit and no double loads?.  */
```
MichielDerhaeg: What's up with all of this punctuation being added?

All of these style changes make it difficult to review.
Can you remove all changes that are not directly relevant to the problem your solving.
  - all the useless punctuation you added.
  - whitespace changes in code and comments
  - indentation changes
It's fine if it is code you touched, but most of it you didn't touch.

I'm currently still reviewing the rest.

gcc/config/arc64/arc64.md
```diff
@@ -968,33 +1313,188 @@ xorl"
...
+       case 7: return "vpack2wl\\t%0,%L1,%H1";
+       case 8: return "addl\\t%0,pcl,%1";
+
+       case 9:  /* Immediate constant store alternative.  */
+
+       /* - Memory Store Alternative (Ustor, r)
+	  If we are on a 32-bit machine core native 64-bit store instructions
+	  do not exist so there is need to extract the base and
+	  calculate offset.  Same happens at case 10 and 11.  */
+       case 13: /* Memory Store Alternative (Ustor, r).  */
```
MichielDerhaeg, line 1407: You need to add fallthrough comments or else the compiler will complain about it.

gcc/config/arc64/arc64.md
```diff
@@ -968,33 +1313,188 @@ xorl"
...
+	  if (!TARGET_64BIT)
+	   {
+	     rtx addr = XEXP (operands[0], 0);
+	     if (GET_CODE (addr) == PLUS && CONST_INT_P (XEXP (addr, 1)))
+	       {
+		 HOST_WIDE_INT offset = INTVAL (XEXP (addr, 1));
+		 const char *src_name = reg_names[REGNO (operands[1])];
```
MichielDerhaeg, line 1414: If case 9 falls through here operand 1 is not a register but an immediate and this will be wrong.

gcc/config/arc64/arc64.md
```diff
@@ -968,33 +1313,188 @@ xorl"
...
+		 const char *base_name = reg_names[REGNO (XEXP (addr, 0))];
+		 if (offset > 0 && (offset % 4 == 0))
+		   {
+		     sprintf (buf, "std.as\t%s,[%s,%ld]", src_name,
+						     base_name, offset/4);
```
MichielDerhaeg, line 1419: Handling of scaled loads and stores should be handled in `arc64_print_operand` with a substitution like %U
That would also prevent needing to use C code here.

gcc/config/arc64/arc64.md
```diff
@@ -968,33 +1313,188 @@ xorl"
...
+		 HOST_WIDE_INT offset = INTVAL (XEXP (addr, 1));
+		 const char *dest =reg_names[REGNO (operands[0])];
+		 const char *base= reg_names[REGNO (XEXP (addr, 0))];
+		 if (offset > 0 && (offset % 4 == 0))
+		   {
+		     sprintf (buf, "ldd.as\t%s,[%s,%ld]", dest, base,offset/4);
+		     return buf;
+		   }
+	       }
+	     return "ldd\\t%0,%1";
```
MichielDerhaeg, line 1445: likewise

gcc/config/arc64/arc64.cc
```diff
@@ -610,20 +639,53 @@ frame_stack_add (HOST_WIDE_INT offset)
...
+  if (!TARGET_LL64 && REG_P (reg) && FP_REGNUM_P (REGNO (reg)))
+    {
+#if 0
+      if (displacement)
+	actual_offset = -displacement;
+      else if (offset == 0)
+	actual_offset = -GET_MODE_SIZE (GET_MODE (reg));
+      else
+	actual_offset = offset - GET_MODE_SIZE (GET_MODE (reg));
+#endif
```
MichielDerhaeg: remove

gcc/config/arc64/arc64.cc
```diff
@@ -5546,9 +5735,19 @@ arc64_expand_prologue (void)
...
+   if (ARC_INTERRUPT_P (cfun->machine->fn_type) && !TARGET_LL64)
+    {
+      if (frame_allocated > 0)
+	{
+	  frame_stack_add ((HOST_WIDE_INT) 0 - frame_allocated);
+	  frame_allocated = 0;
+	}
+    }
+
+  frame_allocated -= arc64_save_callee_saves ();
```
MichielDerhaeg: Doesn't this double allocate space for the callee saves? What's the point of this? Could use documentation. And a test.

gcc/config/arc64/arc64.cc
```diff
@@ -610,20 +639,53 @@ frame_stack_add (HOST_WIDE_INT offset)
...
-   save register REG on stack.  An initial offset OFFSET can be passed
-   to the function.  If a DISPLACEMENT is defined, it will be used to
-   generate pre_modify instead of pre_dec.  */
-
 static HOST_WIDE_INT
 frame_save_reg (rtx reg, HOST_WIDE_INT offset, HOST_WIDE_INT displacement)
 {
   rtx addr, tmp;
 
+  if (!TARGET_LL64 && REG_P (reg) && FP_REGNUM_P (REGNO (reg)))
```
MichielDerhaeg, line 647: All of the stack saving logic you added could use a lot of tests.
If you make mistakes here, it is very likely that run/torture tests just pass even though the stack was allocated incorrectly.

I would explicitly verify the saving & allocating logic.

gcc/config/arc64/arc64.cc
```diff
@@ -1506,6 +1679,14 @@ static unsigned int
...
 			machine_mode mode)
 {
+  if (ARC64_HAS_FP_BASE && FP_REGNUM_P (regno))
+    {
+      if (GET_MODE_SIZE (mode) == 8)
+	return 2;
+      if (GET_MODE_SIZE (mode) == 4)
+	return 1;
+    }
+
```
MichielDerhaeg, line 1761: Don't need to check you're using a 64-bit target here?

gcc/config/arc64/arc64.cc
```diff
@@ -610,20 +639,53 @@ frame_stack_add (HOST_WIDE_INT offset)
...
+      machine_mode mode = GET_MODE (reg);
+      HOST_WIDE_INT mode_size = GET_MODE_SIZE (mode);
+     if (offset)
+	  actual_offset = offset - (displacement ? displacement : mode_size);
+      else if (displacement)
+	  actual_offset = -displacement;
+      else
+	  actual_offset = -mode_size;
+
+      rtx flat_addr = plus_constant (Pmode, stack_pointer_rtx, actual_offset);
```
MichielDerhaeg: wrong formatting

gcc/config/arc64/arc64.cc
```diff
@@ -659,33 +720,74 @@ arc64_save_callee_saves (void)
...
+	{
+	  regno--;
+	  continue;
+	}
 
       save_mode = word_mode;
       if (ARC64_HAS_FP_BASE && FP_REGNUM_P (regno))
 	{
 	  save_mode = ARC64_HAS_FPUD ? DFmode : SFmode;
 	  disp = UNITS_PER_WORD;
```
MichielDerhaeg, line 732: Is `UNITS_PER_WORD` correct here? Shouldn't that depend on `save_mode` or something?

gcc/config/arc64/arc64.cc
```diff
@@ -6721,6 +6920,60 @@ arc64_expand_vector_init (rtx target, rtx vals)
     }
 }
 
+bool
+arc64_insn_has_symbol_p (rtx_insn *insn)
```
MichielDerhaeg, line 7004: You would think there is a generic implementation of this somewhere.

MichielDerhaeg (changes requested):

gcc/config/arc64/arc64.md
```diff
@@ -924,27 +1114,181 @@ xorl"
...
+	emit_move_insn (gpr_hi, mem_hi);  /* ld r3, [r0, 4].  */
+	emit_move_insn (fpu_lo, gpr_lo);  /* mov f0, r2.  */
+	emit_move_insn (fpu_hi, gpr_hi);  /* mov f1, r3.  */
+	rtx_insn *seq = get_insns ();
+	end_sequence ();
+	emit_insn (seq);
+	DONE;
+      }
+
+    rtx hi_dest, hi_src, lo_dest, lo_src;
```
MichielDerhaeg: hardcoding the registers looks very hacky especially if you have to save to the stack manually.

Isn't this just int64_t load + mov to fp reg, can it be implemented that way?

This is a lot of code, please move it to a function in arc64.cc and just call it from here.

MichielDerhaeg: I was thinking: riscv must've solved a similar problem. 
Indeed, rv32d doesn't have fmv.x.d and instead gcc goes through memory.
Check  out movdf_hardfloat_rv32, riscv_split_64bit-move_p and riscv_split_doubleword_move, and riscv_subword.
arm and mips have similar situations that they seem to solve in an entirely different way.

gcc/config/arc64/arc64.cc
Alexehv77: i removed it

gcc/config/arc64/arc64.cc
Alexehv77: done

gcc/config/arc64/arc64.cc
Alexehv77: removed it

gcc/config/arc64/arc64.cc
Alexehv77: My addition that deals with DF is not influenced by this code. As you can see below, inside my code I took care separately of the displacement and at the end i have a continue. So the change to that disp when being influenced by DF is taken care only inside my code.

gcc/config/arc64/arc64.cc
Alexehv77: im now checking for HS6S and depending on this i will make an update

gcc/config/arc64/arc64.cc
Alexehv77: It is needed for interrupts to safeguard the stack frame layout. 
When interrupts happen on a core without 64 bit load/store instructions we cannot rely on incremental stack adjustments during the register saving loop. The code forces the stack pointer (sp) into its final position which is safe. There were tests for which i had to fix this.

gcc/config/arc64/arc64.cc
Alexehv77: is not the same as contains_symbol_ref_p 
inside i check more specific things than contains_symbol_ref_p 
for example:
if (GET_CODE (addr) == SYMBOL_REF || CONST_INT_P (addr))
    return true;
contains_symbol_ref_p will return false;

also this is tailored for what i needed as contains_symbol_ref_p doesnt know about arc 9 bit signed offset
  if (CONST_INT_P (offset))
		    {
		      HOST_WIDE_INT val = INTVAL (offset);
		      if (val < -256 || val > 255)
			return true;
		    }

gcc/config/arc64/arc64.md
Alexehv77: this split happens after reload. the setup is !TARGET_LL64 so no double load/store is possible. I made changes where conditionally for this case a clobber 2 general purpose registers. Otherwise would be possible to search for available registers that are safe to use and in case no registers are found to spill 2 on the stack. For now I will let in place the clobber solution

gcc/config/arc64/arc64.md
Alexehv77: i added

gcc/config/arc64/arc64.md
Alexehv77: ok

gcc/config/arc64/arc64.cc
Alexehv77: there are ~50 lines of code. I made a scrip that took care of my changes and that script touched also the rest of the file. I will push a patch that reverts those changes.

MichielDerhaeg:
hs6x doesn't seem to work:
```
In file included from /home/runner/work/gcc/gcc/arc-gnu-toolchain/newlib/newlib/libc/time/wcsftime.c:68:
  CC       libc/errno/libc_a-errno.o
/home/runner/work/gcc/gcc/arc-gnu-toolchain/newlib/newlib/libc/time/../time/strftime.c: In function '__strftime.isra':
/home/runner/work/gcc/gcc/arc-gnu-toolchain/newlib/newlib/libc/time/../time/strftime.c:1470:1: error: unable to generate reloads for:
 1470 | }
      | ^
(insn 179 183 2810 19 (set (mem/c:SI (plus:DI (reg/f:DI 97 sfp)
                (const_int 1148 [0x47c])) [1 MEM[(wchar_t *)&ctlocbuf + 1020B]+0 S4 A32])
        (const_int 0 [0])) "/home/runner/work/gcc/gcc/arc-gnu-toolchain/newlib/newlib/libc/time/../time/strftime.c":332:28 4 {*arc64_clear_memory}
     (nil))
during RTL pass: reload
/home/runner/work/gcc/gcc/arc-gnu-toolchain/newlib/newlib/libc/time/../time/strftime.c:1470:1: internal compiler error: in curr_insn_transform, at lra-constraints.cc:4355

```

gcc/config/arc64/arc64.cc
MichielDerhaeg: Not done.

gcc/config/arc64/arc64.md
MichielDerhaeg: Not done

gcc/config/arc64/arc64.md
MichielDerhaeg: No?

gcc/config/arc64/arc64.md
MichielDerhaeg: Not fixed? it is checked, but `src_name` will be unitialized?

gcc/config/arc64/arc64.md
MichielDerhaeg: Not done.

gcc/config/arc64/arc64.cc
MichielDerhaeg: Conclusion?

gcc/config/arc64/arc64.cc
MichielDerhaeg: Conclusion?

gcc/config/arc64/arc64.cc
MichielDerhaeg: Then document it.

MichielDerhaeg: > When interrupts happen on a core without 64 bit load/store instructions we cannot rely on incremental stack adjustments during the register saving loop

Why not?

MichielDerhaeg: This specific case will need a test to very the correct offset. E.g. if you accidentally double allocate, a runtime test won't catch that.

gcc/config/arc64/arc64.cc
MichielDerhaeg: This comment still stands. There is no test that explicitly tests stack offsets.
Tests that just contain leaf functions will not trigger this type of code I think.

MichielDerhaeg (changes requested):

gcc/config/arc64/arc64.cc
Alexehv77: the code was fine and it was in sync with what was before.
I now also refactored it a bit:
if (offset)
        actual_offset = offset - (displacement ? displacement : mode_size);
      else if (displacement)
        actual_offset = -displacement;
      else
        actual_offset = -mode_size;
to this:
           HOST_WIDE_INT actual_offset = offset -
              (displacement ? displacement : mode_size);

gcc/config/arc64/arc64.md
Alexehv77: i added   return "stl%U0\\t%1,%0";

gcc/config/arc64/arc64.md
Alexehv77: i have placed  return "stl%U0\\t%1,%0"

gcc/config/arc64/arc64.md
Alexehv77: I will do it in a later patch.

gcc/config/arc64/arc64.md
Alexehv77: the patch was supposed to addresses a fix for a bug.
I propose i do this kind of polishing in a later patch.

gcc/config/arc64/arc64.cc
Alexehv77: The problem is that the 64 bit operations arent atomic and this issue needs to be taken into account in the context of interrupts.
I took out this code and I added a comment regarding the interrupt issue.
I will address this issue in a future patch.

gcc/config/arc64/arc64.cc
Alexehv77: changed it

gcc/config/arc64/arc64.md
Alexehv77: I placed it in the arc.cc

gcc/config/arc64/arc64.cc
Alexehv77: I added a test where this is triggered. 
See frame_save_reg.c

Alexehv77:
> hs6x doesn't seem to work:
> 
> ```
> In file included from /home/runner/work/gcc/gcc/arc-gnu-toolchain/newlib/newlib/libc/time/wcsftime.c:68:
>   CC       libc/errno/libc_a-errno.o
> /home/runner/work/gcc/gcc/arc-gnu-toolchain/newlib/newlib/libc/time/../time/strftime.c: In function '__strftime.isra':
> /home/runner/work/gcc/gcc/arc-gnu-toolchain/newlib/newlib/libc/time/../time/strftime.c:1470:1: error: unable to generate reloads for:
>  1470 | }
>       | ^
> (insn 179 183 2810 19 (set (mem/c:SI (plus:DI (reg/f:DI 97 sfp)
>                 (const_int 1148 [0x47c])) [1 MEM[(wchar_t *)&ctlocbuf + 1020B]+0 S4 A32])
>         (const_int 0 [0])) "/home/runner/work/gcc/gcc/arc-gnu-toolchain/newlib/newlib/libc/time/../time/strftime.c":332:28 4 {*arc64_clear_memory}
>      (nil))
> during RTL pass: reload
> /home/runner/work/gcc/gcc/arc-gnu-toolchain/newlib/newlib/libc/time/../time/strftime.c:1470:1: internal compiler error: in curr_insn_transform, at lra-constraints.cc:4355
> ```

I fixed it
Should pass

PR #254, opened 2026-05-20
fixup! arcv: RMX-500 Limited dual-issue cases

exur00:


exur00:
I don't really know where to put this, it feels like a silly wrapper, but moving the checks inside reorder2 makes it messy and hard to read, moving them to arcv_rmx500_limited_dual_issue_p doesn't feel very right either.

gcc/config/riscv/arcv.cc
```diff
@@ -755,8 +768,10 @@ arcv_sched_reorder2 (rtx_insn **ready, int *n_readyp)
 	  if (NONDEBUG_INSN_P (candidate)
 	      && !SCHED_GROUP_P (candidate)
 	      && (!next_insn || !SCHED_GROUP_P (next_insn))
-	      && arcv_macro_fusion_pair_p
-		   (sched_state.last_scheduled_insn, candidate))
+	      && (arcv_macro_fusion_pair_p
+		   (sched_state.last_scheduled_insn, candidate)
+	      || arcv_macro_limited_dual_issue_pair_p
```
MichielDerhaeg: Add a comment why we check this hear and it isn't part of arcv_marco_fusion_pair_p.

gcc/config/riscv/arcv.cc
```diff
@@ -276,6 +276,29 @@ arcv_rmx500_limited_dual_issue_pair_p (rtx_insn *prev, rtx_insn *curr)
...
+      && arcv_rmx500_limited_dual_issue_pair_p (prev, curr))
+    {
+    if (dump_file)
+      fprintf (dump_file, "ARCV_RMX500_LIMITED_DUAL_ISSUE\n");
+    return true;
+    }
+
+  return false;
+}
+
```
MichielDerhaeg: Let's just get rid of the wrapper. Remove all the checks that aren't necessary. Move the checks that are required to the caller. Perhaps move the fprintf also to the caller.

gcc/config/riscv/arcv.cc
```diff
@@ -276,6 +276,29 @@ arcv_rmx500_limited_dual_issue_pair_p (rtx_insn *prev, rtx_insn *curr)
   return true;
 }
 
+static bool
+arcv_macro_limited_dual_issue_pair_p (rtx_insn *prev, rtx_insn *curr)
+{
+  if (!reload_completed || sched_fusion)
+    return false;
```
MichielDerhaeg: `sched_fusion` is handled by the caller.

gcc/config/riscv/arcv.cc
```diff
@@ -276,6 +276,29 @@ arcv_rmx500_limited_dual_issue_pair_p (rtx_insn *prev, rtx_insn *curr)
...
+
+  if (!TARGET_ARCV_ADVANCED_FUSION)
+    return false;
+
+  if (!single_set (prev) || !single_set (curr))
+    return false;
```
MichielDerhaeg: Handled by the caller?

gcc/config/riscv/arcv.cc
```diff
@@ -276,6 +276,29 @@ arcv_rmx500_limited_dual_issue_pair_p (rtx_insn *prev, rtx_insn *curr)
...
+
+  if (riscv_microarchitecture == arcv_rmx500
```
MichielDerhaeg: Handled by the caller

MichielDerhaeg (changes requested):

MichielDerhaeg (approved):
Let's go.

Issue #253, opened 2026-05-20
gcc from 2603 incorrectly handles cpu_facility


yanovich:
When an object is built with `-mno-code-density`, gcc can still generate an instruction with opcode 0x8024 (`ld_s`) which is from code-density extension. Execution results in instruction error exception, then.

luismgsilva:
Hi. Could you provide more details so we can reproduce the problem? For example:
- target architecture/CPU
- a minimal test case
- the exact command line used to build the object file
- disassembly showing the generated `ld_s` instruction (if available)

yanovich:
Hi,

> Hi. Could you provide more details so we can reproduce the problem? For example:
> 
>     * target architecture/CPU

ARCv2
 
>     * a minimal test case

````
typedef unsigned int uint32_t;

inline uint32_t
readl(uint32_t base, uint32_t offset) {
    return *((volatile uint32_t *)(uint32_t)(base + offset));
}       

int main(void) {
    (void) readl(0x00a30000, 0x0);
    if (readl(0x00a30000, 0x168)) {
            (void) readl(0x00a30000, 0x0);
    };      
}
````


>     * the exact command line used to build the object file

````
$ arc-elf32-gcc -c -Os -mcpu=em4_dmips -mlittle-endian -mno-sdata -mmpy-option=0 -mrf16 -mnorm -mno-code-density -mno-barrel-shifter -Wall -Wextra -Wuninitialized -Werror -Wno-sign-compare -Wno-unused-parameter -Wno-unused-variable -Wno-array-bounds -MD -MP -save-temps=obj -ffunction-sections -fdata-sections -nostdlib -nostartfiles -nostdinc -Wno-unused-label test.c -o test.o
````

>     * disassembly showing the generated `ld_s` instruction (if available)

````
00000000 <main>:
   0:	40c3 00a3 0000      	mov_s	r0,0xa30000
   6:	4104                	ld_s	r1,[r0,0]
   8:	1600 7001 00a3 0168 	ld	r1,[0xa30168]
  10:	e902                	breq_s	r1,0,4	;14 <main+0x14>
  12:	4004                	ld_s	r0,[r0,0]
  14:	700c                	mov_s	r0,0
  16:	7ee0                	j_s	[blink]
````

luismgsilva:
Issue has been fixed and applied on GCC development branch arc-2026.09-dev

luismgsilva:
[...]

PR #255, opened 2026-05-20
arc: Fix code density constraint checks for -mno-code-density

luismgsilva:
Issue originated from https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/issues/253

### DIFF
--- a/gcc/config/arc/constraints.md
+++ b/gcc/config/arc/constraints.md
@@ -292,18 +292,21 @@
   "@internal
    A valid memory operand for ARCompact load instructions"
   (and (match_code "mem")
+       (match_test "!TARGET_V2 || TARGET_CODE_DENSITY")
        (match_test "compact_memory_operand_p (op, mode, false, false)")))
 
 (define_memory_constraint "Uts"
   "@internal
    A valid memory operand for ARCompact load instructions scaled"
   (and (match_code "mem")
-       (match_test "compact_memory_operand_p (op, mode, false, TARGET_CODE_DENSITY)")))
+       (match_test "!TARGET_V2 || TARGET_CODE_DENSITY")
+       (match_test "compact_memory_operand_p (op, mode, false, true)")))
 
 (define_memory_constraint "S"
   "@internal
    A valid memory operand for ARCompact store instructions"
   (and (match_code "mem")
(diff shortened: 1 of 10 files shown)


gcc/testsuite/gcc.target/arc/sdata-6.c
```diff
@@ -28,7 +28,7 @@ TEST (short,   255)
 TEST (int,     255)
 TEST (int64_t, 127)
 
-/* { dg-final { scan-assembler "ldb_s\\s+r0,\\\[gp,@a@sda\\+255\\\]" } } */
```
MichielDerhaeg, line 31: This was wrong?

gcc/config/arc/constraints.md
```diff
@@ -479,7 +483,7 @@
    A valid memory operand for use with code density load ops"
   (and (match_code "mem")
        (match_test "compact_memory_operand_p (op, mode, true, false)")
-       (match_test "TARGET_V2")))
```
MichielDerhaeg, line 482: Not applicable for non TARGET_V2 cores?

gcc/testsuite/gcc.target/arc/sdata-6.c
luismgsilva: Not wrong, but incomplete. The test expected `ldb_s`/`ld_s` but didnt check if code density was available.
Base `mcpu=em` has FLAGS=0 in [arc-cpus.def](https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/blob/arc-2026.09-dev/gcc/config/arc/arc-cpus.def) (no FL_CD), so it generates non-compact instructions by default. The constraints fixes now respect this, so the test needs to accept both forms.

gcc/config/arc/constraints.md
luismgsilva: Yes. `Ucd` is ARCv2-specific. It's for the `av2short` addressing mode (the true parameter in compact_memory_operand_p), which doesn't exist on ARCv1.

ARCv1 cores have their own compact instructions and use the other constraints (`T`, `S`, `Uts`, `Usd`), which now check `!TARGET_V2 || TARGET_CODE_DENSITY` to enable them on both architectures.

MichielDerhaeg (approved):

PR #259, opened 2026-05-21
arcv: FPU latencies for RMX-500

MichielDerhaeg:
This PR also fixed the load latency to 1 cycle. Not sure why it was 2 before.


MichielDerhaeg:
> This PR also fixed the load latency to 1 cycle. Not sure why it was 2 before.

Degrades performance. Let's investigate later..

PR #260, opened 2026-05-21
arcv: check whether uarch allows float memory fusions.

exur00:


gcc/config/riscv/arcv.cc
MichielDerhaeg: these also?

gcc/config/riscv/arcv.cc
```diff
@@ -954,7 +971,11 @@ arcv_sched_adjust_priority (rtx_insn *insn, int priority)
       && ((get_attr_type (insn) == TYPE_STORE
 	   && get_attr_type (next) == TYPE_STORE)
 	 || (get_attr_type (insn) == TYPE_LOAD
-	     && get_attr_type (next) == TYPE_LOAD)))
+	     && get_attr_type (next) == TYPE_LOAD)
+	 || (get_attr_type (insn) == TYPE_FPLOAD
+	     && get_attr_type (next) == TYPE_FPLOAD)
+	 || (get_attr_type (insn) == TYPE_FPSTORE
+	     && get_attr_type (next) == TYPE_FPSTORE)))
```
MichielDerhaeg: also check rmx500 here?

gcc/config/riscv/arcv.cc
```diff
@@ -1148,7 +1169,7 @@ arcv_sched_variable_issue (rtx_insn *insn, int more)
 
   if (next && SCHED_GROUP_P (next))
     {
-      if (arcv_memop_p (insn) || arcv_memop_p (next))
+      if (arcv_fusible_memop_p (insn) || arcv_fusible_memop_p (next))
```
MichielDerhaeg: technically it doesn't matter if it is fusible, if it is a float load or store, the b pipeline will be filled.

gcc/config/riscv/arcv.cc
```diff
@@ -822,7 +839,7 @@ arcv_sched_reorder2 (rtx_insn **ready, int *n_readyp)
   if (!sched_state.pipeB_scheduled_p && sched_state.last_scheduled_insn
       && ready && *n_readyp > 0
       && !SCHED_GROUP_P (sched_state.last_scheduled_insn)
-      && arcv_memop_p (sched_state.last_scheduled_insn))
+      && arcv_fusible_memop_p (sched_state.last_scheduled_insn))
```
MichielDerhaeg: Don't think this is correct. floating-point mem-ops still fill the b pipe. This is now avoided.
Past this point is all RHX code. Technically arcv_memop_p is correct here I think.

MichielDerhaeg (changes requested):

MichielDerhaeg (approved):

PR #261, opened 2026-05-26
Add matrix address optimization phase for Zba

Alexehv77:
This pass identifies addressing patterns in memory-access intensive kernels. In such kernels an index is frequently scaled and added to a base pointer to access memory followed by an immediate index update. The pass hoists the pointer arithmetic out of the loop header converting index addressing into induction variables.

### DIFF
--- a/gcc/config/riscv/riscv-matrix-opt.cc
+++ b/gcc/config/riscv/riscv-matrix-opt.cc
@@ -1,0 +1,364 @@
+/* This pass identifies addressing patterns in memory-access intensive 
+   kernels. In such kernels an index is frequently scaled and added to a 
+   base pointer to access memory followed by an immediate index update.
+   The pass hoists the pointer arithmetic out of the loop header converting
+   index addressing into induction variables. */
+
+#include "config.h"
+#include "system.h"
+#include "coretypes.h"
+#include "backend.h"
+#include "target.h"
+#include "rtl.h"
+#include "tree.h"
+#include "gimple.h"
+#include "cfghooks.h"
+#include "cfgcleanup.h"
+#include "memmodel.h"
+#include "emit-rtl.h"
+#include "alias.h"
+#include "rtlanal.h"
+#include "cfgloop.h"
+#include "insn-config.h"
+#include "insn-flags.h"
+#include "insn-codes.h"
+#include "optabs.h"
+#include "recog.h"
+#include "tree-pass.h"
+#include "cfgrtl.h"
+#include "expr.h"
+#include "explow.h"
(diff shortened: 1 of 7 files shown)


gcc/config/riscv/riscv-matrix-opt.cc
```diff
@@ -0,0 +1,364 @@
+/* This pass identifies addressing patterns in memory-access intensive 
+   kernels. In such kernels an index is frequently scaled and added to a 
+   base pointer to access memory followed by an immediate index update.
+   The pass hoists the pointer arithmetic out of the loop header converting
+   index addressing into induction variables. */
+
```
MichielDerhaeg, line 6: On https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/241 I asked what the point of this pass was. This looks a lot like something ivopts is for. Why make something specialized for RTL?

Alexehv77: In the context of loops with unsigned loop bounds, ivopts can not do much because scalar evolution can not calculate the upper iteration bound of a loop - as the loop iterator may overflow. Because of this i tried to fix things at rtl level.

As you were trying to make use of loop versioning at the tree level (where ivopts takes place) I started approaching the performance issues from the rtl level by analyzing the generated assembly and comparing it to what llvm and gcc with signed loop bounds were generating.

This phase combined with the patch from https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/261 managed to reach same performance as if the loop bounds were signed int.

Also remark that ivopts is not aware of sh1add.uw and addw and therefore its cost model can not accurately calculate whether hoisting them is cheaper than leaving them inside the loop.

In conclusion this phase is complementary to ivopts as it targets a specific peephole (or combine) type of optimization being able to reason without the issues triggered by the unsigned loop bounds.
Remark here that combine and peephole wont apply here as they work on instruction windows of 2-4 instructions while the pattern i recognize here is a dependency chain of several instructions that spans over bb or even loop boundaries. 

Also another interesting issue is that i allow the optimization to take place only for leaf functions (this ensures no function calls happen inside - and as a result no need to spill caller saved registers) to decrease the chance of introducing spill code. On the contrary i am familiar that although ivopts (as running at the tree level) does take register pressure into account it does it in simplistic way which in return has the potential to result  in a large amount of spill code.

Last but not least - there is a jira ticket where i proposed another solution for dealing with loops with unsigned bounds that doesnt require loop versioning. The approach is rather simple and i have already implemented a basic tree level phase that deals with it.

PR #263, opened 2026-06-10
arcv: check_effective_target_arcv_* for other default ABI's

MichielDerhaeg:
These checks failed in case e.g. -mabi=ilp32d was configured to be the default because the test fails for march=rv32im* because of an ABI mismatch if no D extension is specified.

This caused xarcv extension tests to silently be skipped.

### DIFF
--- a/gcc/testsuite/lib/target-supports.exp
+++ b/gcc/testsuite/lib/target-supports.exp
@@ -14137,12 +14137,13 @@ proc check_effective_target_arcv_udsp { } {
     if { !([istarget riscv32-*-*]) } {
          return 0
      }
+    set gcc_march [riscv_get_arch]
     return [check_no_compiler_messages arcv_udsp object {
         void foo (void)
         {
           asm ("arcv.xvadd.vv x0, x0, x0, e16, m1");
         }
-    } "-march=rv32i_xarcvudsp" ]
+    } "-march=${gcc_march}_xarcvudsp" ]
 }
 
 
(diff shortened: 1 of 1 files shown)


gcc/testsuite/lib/target-supports.exp
```diff
@@ -14137,12 +14137,13 @@ proc check_effective_target_arcv_udsp { } {
     if { !([istarget riscv32-*-*]) } {
          return 0
      }
+    set gcc_march [riscv_get_arch]
```
exur00, line 14140: Why is it here this simple

gcc/testsuite/lib/target-supports.exp
```diff
@@ -14152,96 +14153,104 @@ proc check_effective_target_arcv_bitrev { } {
...
           }
-  } "-march=rv32im_xarcvbitstream" ]
+  } "-march=${gcc_march}_xarcvbitstream" ]
 }
 # Return 1 if the ARC-V VDSP extension is available.
 proc check_effective_target_arcv_vdsp { } {
     if { !([istarget riscv*-*-*]) } {
         return 0
     }
+    set gcc_march [regsub {[[:alnum:]]*} [riscv_get_arch] &v]
```
exur00, line 14182: But here more compilex? wouldn't the same work here?

gcc/testsuite/lib/target-supports.exp
MichielDerhaeg: These extensions don't rely on RVV.

gcc/testsuite/lib/target-supports.exp
MichielDerhaeg: This needs the right RVV extension in addition to the default.

MichielDerhaeg:
I will merge if no more remarks today.

luismgsilva (approved):
Looks good to me

exur00 (approved):
Looks good

PR #265, opened 2026-06-11
Add GitHub CI workflows for selected targets

MichielDerhaeg:


MichielDerhaeg:
Baseline comes from trunk, needs to be updated once the builds finish

PR #270, opened 2026-06-16
arc: fix for extvsi split patterns

Alexehv77:
Description change to the pattern define_insn_and_split "*extvsi_n_0"

If using && reload_completed the compiler is forced to wait until after the RA/Reload pass is entirely finished. Physical registers are already permanently locked down. If using && 1 the split is allowed to occur early such as during the combine pass when registers are still flexible pseudos. Splitting earlier (using && 1) is generally better than waiting until after reload—provided as it provides significant optimization advantages.

Description change to pattern define_insn_and_split "*extvsi_1_0" keeping "0" with && 1 is not in accordance with the GCC instruction pattern rules:

1. According to RTL instruction definition by writing "0" we are telling the dataflow analysis that operand 0 and operand 1 must share the exact same physical register. By writing && 1 we are telling the compiler to split this instruction into pieces during the very early optimization passes (like combine) which is before physical registers are even handed out the "0" constraint.

2. When you split an instruction early using && 1 the purpose is to allow subsequent optimization passes (like Scheduling for register pressure, cse, dce etc) to optimize the broken instructions. If you combine an early split with a restrictive "0" the df-scan gets confused about the exact life ranges of the pseudo registers. Because it thinks the destination and source are linked it may assume a register becomes dead earlier than it actually does. This prevents subsequent passes from safely reordering instructions or eliminating redundant operations resulting in less optimized code.

### DIFF
--- a/gcc/config/arc/arc.md
+++ b/gcc/config/arc/arc.md
@@ -6268,7 +6268,7 @@ archs4x, archs4xd"
 ;; Split sign-extension of single least significant bit as and x,$1;neg x
 (define_insn_and_split "*extvsi_1_0"
   [(set (match_operand:SI 0 "register_operand" "=r")
-	(sign_extract:SI (match_operand:SI 1 "register_operand" "0")
+	(sign_extract:SI (match_operand:SI 1 "register_operand" "r")
 			 (const_int 1)
 			 (const_int 0)))]
   "!TARGET_BARREL_SHIFTER"
@@ -6288,7 +6288,7 @@ archs4x, archs4xd"
    && IN_RANGE (INTVAL (operands[2]), 2,
 		(optimize_insn_for_size_p () ? 28 : 30))"
   "#"
-  "&& reload_completed"
+  "&& 1"
 [(set (match_dup 0) (and:SI (match_dup 1) (match_dup 3)))
  (set (match_dup 0) (xor:SI (match_dup 0) (match_dup 4)))
  (set (match_dup 0) (minus:SI (match_dup 0) (match_dup 4)))]
--- a/gcc/testsuite/gcc.target/arc/extvsi-sign-extract.c
+++ b/gcc/testsuite/gcc.target/arc/extvsi-sign-extract.c
@@ -1,0 +1,67 @@
+/* { dg-do run } */
+/* { dg-options "-O2 -mcpu=em -save-temps" } */
+
+#include <stdio.h>
+#include <stdlib.h>
+
+/* Test multi-bit Extraction Pattern (*extvsi_n_0).  */
+struct bitstruct {
+  unsigned a : 21;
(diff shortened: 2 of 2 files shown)


gcc/config/arc/arc.md
```diff
@@ -6268,7 +6268,7 @@ archs4x, archs4xd"
 ;; Split sign-extension of single least significant bit as and x,$1;neg x
 (define_insn_and_split "*extvsi_1_0"
   [(set (match_operand:SI 0 "register_operand" "=r")
-	(sign_extract:SI (match_operand:SI 1 "register_operand" "0")
+	(sign_extract:SI (match_operand:SI 1 "register_operand" "r")
```
MichielDerhaeg, line 6271: I was under the impression that this was a functional issue.
But the resulting RTL pattern is correct?

This is just for performance?

Alexehv77: The first define_insn_and_split has potential to generate incorrect code.

For the second one there was a change from && 1 to &&reload_completed and at the same time there was a change from (sign_extract:SI (match_operand:SI 1 "register_operand" "0") 
to                  (sign_extract:SI (match_operand:SI 1 "register_operand" "r").

My point is that there is no need to have the both as the second change (i.e,  (sign_extract:SI (match_operand:SI 1 "register_operand" "0") to (sign_extract:SI (match_operand:SI 1 "register_operand" "r")) is enough allowing the split to take place even before ra.

MichielDerhaeg:
I'm inclined not to go through the effort of arguing to get this merged upstream unless it solves a real functional issue.
It is simply not worth the effort.

> According to RTL instruction definition by writing "0" we are telling the dataflow analysis that operand 0 and operand 1 must share the exact same physical register.

df-scan doesn't read the constraints, just checks which registers are filled in the operands in RTL, which are not enforced to be the same before RA. They are, in fact, different registers in the test you made. After RA they will be the same, which might also happen after your change by chance, don't see a problem there.

By grepping GCC i found that constraints are only read here:
  ira.cc      reload.cc     reload1.cc    lra-constraints.cc
  caller-save.cc   postreload-gcse.cc  ree.cc           reorg.cc (post-reload)
Which means that when the split happens before reload, the register constraints don't have any effect. So why bother?

Also the test doesn't fail *before* the changes in arc.md, so I don't see the point of that either.

So again, if this is not a functional issue, let's not bother with this. Don't think this is worth the effort, and let's drop it.
And that includes https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/269 because I don't want downstream to deviate from upstream.

MichielDerhaeg:
I think we agreed to keep it simple and push it upstream as an non-functional change.
Can you update this?
Keep the commit message very short and simple. And I would drop the test as it is fixing a functional issue.

PR #271, opened 2026-06-16
arc: Do not pass -mcode-density to the assembler with -mno-code-density

luismgsilva:
arc: Do not pass -mcode-density to the assembler with -mno-code-density
    
    For ARC EM CPUs that enable code density by default, cpu_to_as adds
    -mcode-density to the assembler command line.  When the user passes
    -mno-code-density, the compiler disables the extension, but cpu_to_as
    still requested -mcode-density for those CPUs.  The GNU assembler only
    has a positive -mcode-density option and no way to disable the
    extension from the command line, so the driver must omit -mcode-density
    instead of forwarding the user's -mno-code-density flag.
    
    Pass a marker from ASM_SPEC when -mno-code-density is given, and
    suppress the default -mcode-density request in cpu_to_as.  Explicit
    -mcode-density continues to be forwarded by %{mcode-density}.
    
    For example, with -mcpu=em4_dmips the driver invokes the assembler as:
    
            gcc ... -mcpu=em4_dmips -c test.c
            -> as ... -mcpu=arcem -mcode-density ...
    
    With -mno-code-density, it used to make no difference:
    
            gcc ... -mcpu=em4_dmips -mno-code-density -c test.c
            -> as ... -mcpu=arcem -mcode-density ...
    
    and now correctly omits -mcode-density:
    
            gcc ... -mcpu=em4_dmips -mno-code-density -c test.c
            -> as ... -mcpu=arcem ...
    
    Assembling the same test program shows different instruction encodings
    in the object file, for example:
    
            $ gcc -c test.c -mcpu=em4_dmips
            $ objdump -d test.o
               4:   4104                    ld_s    r1,[r0,0]
              10:   4004                    ld_s    r0,[r0,0]
    
    versus:
    
            $ gcc -c test.c -mcpu=em4_dmips -mno-code-density
            $ objdump -d test.o
               4:   8020                    ld_s    r1,[r0,0]
              10:   8000                    ld_s    r0,[r0,0]
    
    gcc/ChangeLog:
    
            * config/arc/arc.h (ASM_SPEC): Pass a no-cd marker to cpu_to_as
            when -mno-code-density is given.
            * config/arc/driver-arc.cc (arc_cpu_to_as): Do not request
            -mcode-density from the assembler when the marker is present.
    
    Signed-off-by: Luis Silva <luis.silva@globalfoundries.com>

### DIFF
--- a/gcc/config/arc/driver-arc.cc
+++ b/gcc/config/arc/driver-arc.cc
@@ -25,6 +25,25 @@
 #include "coretypes.h"
 #include "tm.h"
 
+/* Look up the CPU given by NAME, or the default CPU if NAME is NULL.  */
+
+static const arc_cpu_t *
+arc_lookup_cpu (const char *name)
+{
+  if (name == NULL)
+    return &arc_cpu_types[(int) TARGET_CPU_DEFAULT];
+
+  for (const arc_cpu_t *arc_selected_cpu = arc_cpu_types;
+       arc_selected_cpu->name;
+       arc_selected_cpu++)
+    {
+      if (strcmp (arc_selected_cpu->name, name) == 0)
+	return arc_selected_cpu;
+    }
+
+  gcc_unreachable ();
+}
+
 /* Returns command line parameters to pass to as.  */
 
 const char*
@@ -33,29 +52,12 @@ arc_cpu_to_as (int argc, const char **argv)
   const char *name = NULL;
   const arc_cpu_t *arc_selected_cpu;
 
-  /* No argument, check what is the default cpu.  */
-  if (argc == 0)
-    {
-      arc_selected_cpu = &arc_cpu_types[(int) TARGET_CPU_DEFAULT];
-    }
-  else
-    {
-      name = argv[0];
-      for (arc_selected_cpu = arc_cpu_types; arc_selected_cpu->name;
-	   arc_selected_cpu++)
-	{
-	  if (strcmp (arc_selected_cpu->name, name) == 0)
-	    break;
-	}
-    }
+  arc_selected_cpu = arc_lookup_cpu (argc == 0 ? NULL : argv[0]);
 
   switch (arc_selected_cpu->arch_info->arch_id)
     {
     case BASE_ARCH_em:
-      if (arc_selected_cpu->flags & FL_CD)
-	name = "-mcode-density";
-      else
-	name = "";
+      name = "";
       if (arc_selected_cpu->flags & FL_FPUDA)
 	name = concat ("-mfpuda ", name, NULL);
       if (arc_selected_cpu->flags & FL_SPFP)
@@ -80,4 +82,30 @@ arc_cpu_to_as (int argc, const char **argv)
       gcc_unreachable ();
     }
   return NULL;
+}
+
+/* Returns code density parameters to pass to as.  */
+
+const char *
+arc_cd_to_as (int argc, const char **argv)
+{
+  const char *name = NULL;
+  const arc_cpu_t *arc_selected_cpu;
+
+  for (int i = 0; i < argc; i++)
+    {
+      if (strcmp (argv[i], "no-cd") == 0)
+	return "";
+      else if (strcmp (argv[i], "cd") == 0)
+	return "-mcode-density";
+      else
+	name = argv[i];
+    }
+
+  arc_selected_cpu = arc_lookup_cpu (name);
+  if (arc_selected_cpu->arch_info->arch_id == BASE_ARCH_em
+      && (arc_selected_cpu->flags & FL_CD))
+    return "-mcode-density";
+
+  return "";
 }
--- a/gcc/config/arc/arc.h
+++ b/gcc/config/arc/arc.h
@@ -70,9 +70,11 @@ along with GCC; see the file COPYING3.  If not see
 %{G*}								\
 "
 extern const char *arc_cpu_to_as (int argc, const char **argv);
+extern const char *arc_cd_to_as (int argc, const char **argv);
 
 #define EXTRA_SPEC_FUNCTIONS			\
-  { "cpu_to_as", arc_cpu_to_as },
+  { "cpu_to_as", arc_cpu_to_as },		\
+  { "cd_to_as", arc_cd_to_as },
 
 /* This macro defines names of additional specifications to put in the specs
    that can be used in various specifications like CC1_SPEC.  Its definition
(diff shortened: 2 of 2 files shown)


luismgsilva:
This patch fixes the original issue reported in https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/issues/253

luismgsilva:
@MichielDerhaeg I split the code-density logic into `arc_cd_to_as`, separate from `arc_cpu_to_as`, as we discussed previously (128812682491aa490879f26b33a9c06cc11ec3cd).

`cd_to_as` still needs the CPU name from `-mcpu=...` (or the gcc configure cpu) so it can decide whether the assembler gets `-mcode-density` by default.  For EM CPUs with `FL_CD` (e.g., em4_dmips), we pass it unless the user passes `-mno-code-density`.

gcc/config/arc/driver-arc.cc
```diff
@@ -81,3 +83,29 @@ arc_cpu_to_as (int argc, const char **argv)
...
+{
+  const char *name = NULL;
+  const arc_cpu_t *arc_selected_cpu;
+
+  for (int i = 0; i < argc; i++)
+    {
+      if (strcmp (argv[i], "no-cd") == 0)
+	return "";
+      else if (strcmp (argv[i], "cd") == 0)
+  return "-mcode-density";
```
MichielDerhaeg: what happened to the formatting here? :D

gcc/config/arc/driver-arc.cc
```diff
@@ -25,37 +25,39 @@
...
+
+  for (arc_selected_cpu = arc_cpu_types; arc_selected_cpu->name;
+       arc_selected_cpu++)
+    {
+      if (strcmp (arc_selected_cpu->name, name) == 0)
+	break;
+    }
+  return arc_selected_cpu;
+}
+
```
MichielDerhaeg, line 46: And what if the cpu is not found? can this happen?

luismgsilva: This means the CPU specified by the user is invalid. As a result, execution will never reach this part of the code because the compiler fails earlier with the following error:

```
arc-snps-elf-gcc: error: unrecognized argument in option '-mcpu=test'
arc-snps-elf-gcc: note: valid arguments to '-mcpu=' are: arc600 arc600_mul32x16 arc600_mul64 arc600_norm arc601 arc601_mul32x16 arc601_mul64 arc601_norm arc700 arcem archs em em4 em4_dmips em4_fpuda em4_fpus em_mini hs hs34 hs38 hs38_linux hs4x hs4x_rel31 hs4xd nps400 quarkse_em
```

Either way we can add a `gcc_unreachable` after the loop.

```suggestion
static const arc_cpu_t *
arc_lookup_cpu (const char *name)
{
  const arc_cpu_t *arc_selected_cpu;

  if (name == NULL)
    return &arc_cpu_types[(int) TARGET_CPU_DEFAULT];

  for (arc_selected_cpu = arc_cpu_types; arc_selected_cpu->name;
       arc_selected_cpu++)
    {
      if (strcmp (arc_selected_cpu->name, name) == 0)
	return arc_selected_cpu;
    }
  gcc_unreachable ();
}

```

gcc/config/arc/driver-arc.cc
luismgsilva: Opss

gcc/config/arc/driver-arc.cc
MichielDerhaeg: I think that's clearer, perhaps also declare `arc_selected_cpu` inside the `for (`

luismgsilva: Done

MichielDerhaeg (approved):

luismgsilva:
Patch sent upstream: https://gcc.gnu.org/pipermail/gcc-patches/2026-July/723879.html

luismgsilva:
Added test case as per Claudiu's request https://gcc.gnu.org/pipermail/gcc-patches/2026-September/731628.html

gcc/testsuite/gcc.target/arc/mno-code-density-flag.c
```diff
@@ -0,0 +1,12 @@
+/* { dg-do assemble } */
+/* { dg-require-effective-target codedensity } */
+/* { dg-skip-if "" { ! { arcem } } } */
+/* { dg-options "-mno-code-density" } */
```
MichielDerhaeg, line 4: `dg-require-effective-target` is not useful if you force `arcem`.

nit; The `code-density-flag.c` and many others use this pattern:
```
{ dg-skip-if "" { ! { clmcpu } } }
{ dg-options "-mcpu=em_mini -mcode-density" }
```

luismgsilva: > dg-require-effective-target is not useful if you force arcem.

Indeed.

> nit; The code-density-flag.c and many others use this pattern:
> { dg-skip-if "" { ! { clmcpu } } }
> { dg-options "-mcpu=em_mini -mcode-density" }

Yes, but my point was to be able to run the test for all available EM CPUs. For example, with this approach, if `-mcpu=em4` is passed and the test case is defined with `-mcpu=em4_dmips`, the test will be skipped, even though it would normally pass.

luismgsilva: > dg-require-effective-target is not useful if you force arcem.

Actually, for `-mcpu=em`, it would always be true because `em` does not enable code density by default. The purpose of `dg-require-effective-target` is to check whether code density is enabled by default.

MichielDerhaeg: If you do
```
{ dg-skip-if "" { ! { clmcpu } } }
{ dg-options "-mcpu=arcem -mno-code-density" }
```
The test will always run, and correctly. And not depend on whatever the default cpu the toolchain is configured with.

luismgsilva: > If you do
> 
> ```
> { dg-skip-if "" { ! { clmcpu } } }
> { dg-options "-mcpu=arcem -mno-code-density" }
> ```
> 
> The test will always run, and correctly. And not depend on whatever the default cpu the toolchain is configured with.

Right, but AFAIK jenkins always passes the mcpu so this test would always be skipped

MichielDerhaeg: But I suggested this because that's what the other tests do, do we always skip all of these tests?

luismgsilva: Well, don't know if its related to the mcpu flag, but:
```
$ grep -r code-density-flag.c | grep sum
archs/gcc.sum:UNSUPPORTED: gcc.target/arc/code-density-flag.c
em/gcc.sum:UNSUPPORTED: gcc.target/arc/code-density-flag.c
arc600/gcc.sum:UNSUPPORTED: gcc.target/arc/code-density-flag.c
em4_fpuda/gcc.sum:UNSUPPORTED: gcc.target/arc/code-density-flag.c
hs34/gcc.sum:UNSUPPORTED: gcc.target/arc/code-density-flag.c
arcem/gcc.sum:UNSUPPORTED: gcc.target/arc/code-density-flag.c
hs4xd/gcc.sum:UNSUPPORTED: gcc.target/arc/code-density-flag.c
hs38_linux/gcc.sum:UNSUPPORTED: gcc.target/arc/code-density-flag.c
arc700/gcc.sum:UNSUPPORTED: gcc.target/arc/code-density-flag.c
hs/gcc.sum:UNSUPPORTED: gcc.target/arc/code-density-flag.c
```

MichielDerhaeg: ok conclusion: clmcpu = bad; arcem requirement needs to be dropped because it overlaps with require_effective_target?

luismgsilva: If you remove:
- `/* { dg-skip-if "" { ! { arcem } } } */` it will also execute for `hs` family which will result in a error because you cannot disable code-density.
- `/* { dg-require-effective-target codedensity } */` it will always pass on `-mcpu=em` because code-density is already disabled by default (which is not that big of a deal tbh).

PR #273, opened 2026-06-17
arcv: Use correct function to get next fusible insn

MichielDerhaeg:
We noticed that whatever arcv_next_fusible_insn did, did not entirely
correspond to how the scheduler pairs instructions. The scheduler uses
next_nonnote_nondebug_insn_bb so we should as well.

In particular, `sched_macro_fuse_insns (insn)` uses

	prev = prev_nonnote_nondebug_insn_bb

and checks

	prev
	&& (any_condjump_p (insn)
	    || single_set (insn) && single_set (insn))

arcv_macro_fusion_pair_p already checks for all of that. So if we don't
use this function we check these manually before calling
get_attr_type or arcv_memop_p. The following expression tends to capture
all valid cases:

	recog_memoized (insn) >= 0

### DIFF
--- a/gcc/config/riscv/arcv.cc
+++ b/gcc/config/riscv/arcv.cc
@@ -75,30 +75,6 @@ struct arcv_sched_state {
 
 static struct arcv_sched_state sched_state;
 
-/* Return the next possible fusible insn.  */
-
-static rtx_insn *
-arcv_next_fusible_insn (rtx_insn *insn)
-{
-  while (insn)
-    {
-      insn = NEXT_INSN (insn);
-
-      if (insn == 0)
-	break;
-
-      if (NOTE_INSN_BASIC_BLOCK_P (insn) || JUMP_TABLE_DATA_P (insn))
-  return NULL;
-
-      if (!NONDEBUG_INSN_P (insn) || GET_CODE (PATTERN (insn)) == USE)
-	continue;
-
-  break;
-    }
-
-  return insn;
-}
-
 /* Return TRUE if the target microarchitecture supports macro-op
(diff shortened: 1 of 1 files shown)


exur00 (approved):
Nice, seems like a perfect replacement.

MichielDerhaeg:
> Nice, seems like a perfect replacement.

Doesn't work though, needs fixing :p

MichielDerhaeg:
> > Nice, seems like a perfect replacement.
> 
> Doesn't work though, needs fixing :p

This should work now, please check again.
Some TODO's left though as you can see.

gcc/config/riscv/arcv.cc
```diff
@@ -1158,14 +1138,15 @@ arcv_can_issue_more_p (int issue_rate, int more)
...
-  rtx_insn *next = arcv_next_fusible_insn (insn);
+  rtx_insn *next = next_nonnote_nondebug_insn_bb (insn);
 
   if (riscv_microarchitecture == arcv_rmx500)
     {
       sched_state.last_scheduled_insn = insn;
 
-      if (get_attr_type (insn) == TYPE_ALU_FUSED
-	  || get_attr_type (insn) == TYPE_IMUL_FUSED)
+      if (single_set (insn)
```
MichielDerhaeg, line 1145: Are instrunctions with clobbers single_set's?

MichielDerhaeg:
This was validated with `--enable-checking=rtl`. Note that this will still fail, but those issues are not related to the pattern I tried to fix here. The other fixes will be fixed in a separate PR.

PR #274, opened 2026-06-24
arc: Fix incorrect RTL generation for store instruction

Alexehv77:
arc: Fix invalid register index addressing for uncached stores

For uncached stores (direct memory stores) the compiler was generating indexed addressing 
using register offsets. However, the arc target doesnt support such stores.

This was happening becasue prepare_move_operands() emits a raw VUNSPEC_ARC_STDI
and returns true - bypassing in this was further validation which was for example happening 
for standard stores. This was fised during rtl expand by validating uncached stores against 
move_dest_operand() and forcing complex addresses into separate register (via copy_to_mode_reg()).

The issue was originally described here:
https://github.com/foss-for-synopsys-dwc-arc-processors/toolchain/issues/672
where the store instructions (like stb.di) were emitting invalid
memory offsets.

Also added dejagnu tests to verify that the correct assembly instruction and offsets are 
generated in the output file.

### DIFF
--- a/gcc/config/arc/arc.md
+++ b/gcc/config/arc/arc.md
@@ -4954,8 +4954,9 @@ archs4x, archs4xd"
   "ld<mALLI>%U1.di\\t%0,%1"
   [(set_attr "type" "load")])
 
+; Direct store pattern
 (define_insn "stdi<mode>"
-  [(unspec_volatile [(match_operand:ALLI 0 "memory_operand"    "m,m,Usc")
+  [(unspec_volatile [(match_operand:ALLI 0 "move_dest_operand" "m,m,Usc")
 		     (match_operand:ALLI 1 "nonmemory_operand" "r,Cm3,i")]
 		    VUNSPEC_ARC_STDI)]
   ""
@@ -4964,7 +4965,7 @@ archs4x, archs4xd"
    (set_attr "type" "store")])
 
 (define_insn_and_split "*stdidi_split"
-  [(unspec_volatile [(match_operand:DI 0 "memory_operand"   "m")
+  [(unspec_volatile [(match_operand:DI 0 "non_incdec_memory_operand"   "m")
 		     (match_operand:DI 1 "register_operand" "r")]
 		    VUNSPEC_ARC_STDI)]
   "!TARGET_LL64"
@@ -4980,7 +4981,7 @@ archs4x, archs4xd"
    operands[4] = gen_highpart (SImode, operands[0]);
   }
   "
-  )
+  [(set_attr "type" "store")])
 
 (define_insn_and_split "*lddidi_split"
   [(set (match_operand:DI 0 "register_operand" "=r")
@@ -4999,6 +5000,7 @@ archs4x, archs4xd"
    operands[4] = gen_highpart (SImode, operands[0]);
   }
   "
+  [(set_attr "type" "load")]
   )
 
 (define_insn "trap_s"
--- a/gcc/config/arc/predicates.md
+++ b/gcc/config/arc/predicates.md
@@ -775,3 +775,22 @@
   (ior (match_test "register_operand (op, mode)")
        (and (match_code "const_int, symbol_ref")
 	    (match_test "!optimize_size"))))
+
+(define_predicate "non_incdec_memory_operand"
+  (match_operand 0 "move_dest_operand")
+{
+  rtx addr = XEXP (op, 0);
+  enum rtx_code code = GET_CODE (addr);
+
+  /* Reject post/pre inc/dec to be used for DI mode on
+     targets that dont have 64 ls units.  */
+  if (code == POST_INC
+      || code == PRE_INC
+      || code == POST_DEC
+      || code == PRE_DEC
+      || code == POST_MODIFY
+      || code == PRE_MODIFY)
+    return false;
+
+  return true;
+})
--- a/gcc/testsuite/gcc.target/arc/uncached-stdi.h
+++ b/gcc/testsuite/gcc.target/arc/uncached-stdi.h
@@ -1,0 +1,43 @@
+#ifndef __uncached
+#define __uncached __attribute__((uncached))
+#endif
+
+/*Test for QImode.  */
+volatile char *
+pack_byte_ptr(volatile char *p, const void *s, int len)
+{
+    volatile __uncached char *buf = p + 4;
+    for (int i = 0; i < len; i++)
+        buf[i] = ((const char *)s)[i];
+    return p;
+}
+
+/*Test for HImode.  */
+volatile short *
+pack_halfword_ptr(volatile short *p, const void *s, int len)
+{
+    volatile __uncached short *buf = p + 4;
+    for (int i = 0; i < len; i++)
+        buf[i] = ((const short *)s)[i];
+    return p;
+}
+
+/*Test for SImode.  */
+volatile int *
+pack_word_ptr(volatile int *p, const void *s, int len)
+{
+    volatile __uncached int *buf = p + 4;
+    for (int i = 0; i < len; i++)
+        buf[i] = ((const int *)s)[i];
+    return p;
+}
+
+/*Test for DImode.  */
+volatile long long *
+pack_doubleword_ptr(volatile long long *p, const void *s, int len)
+{
+    volatile __uncached long long *buf = p + 4;
+    for (int i = 0; i < len; i++)
+        buf[i] = ((const long long *)s)[i];
+    return p;
+}
--- a/gcc/config/arc/arc.cc
+++ b/gcc/config/arc/arc.cc
@@ -9282,16 +9282,33 @@ prepare_move_operands (rtx *operands, machine_mode mode)
 	      return true;
 	    }
 	}
-      /* Second, we check for the uncached.  */
+
+      /* Second, handle uncached stores.  */
       if (arc_is_uncached_mem_p (operands[0]))
 	{
-	  if (!REG_P (operands[1]))
-	    operands[1] = force_reg (mode, operands[1]);
+	   if (!REG_P (operands[1]))
+	   operands[1] = force_reg (mode, operands[1]);
+
+	   /* Direct uncached stores (st.di) dont support register index
+	      [reg + reg] addressing.  Check and fix the address before
+	      emitting VUNSPEC_ARC_STDI.  */
+	  if (!move_dest_operand (operands[0], mode))
+	    {
+	      /* Force the [reg + reg] in the base register as a separate
+		 insn.  */
+	      rtx tmp0 = copy_to_mode_reg (Pmode, XEXP (operands[0], 0));
+	      rtx tmp1 = change_address (operands[0], mode, tmp0);
+	      MEM_COPY_ATTRIBUTES (tmp1, operands[0]);
+	      operands[0] = tmp1;
+	    }
+
 	  emit_insn (gen_rtx_UNSPEC_VOLATILE
 		     (VOIDmode, gen_rtvec (2, operands[0], operands[1]),
 		      VUNSPEC_ARC_STDI));
 	  return true;
 	}
+
+      /* Third, handle uncached loads.  */
       if (arc_is_uncached_mem_p (operands[1]))
 	{
 	  rtx tmp = operands[0];
--- a/gcc/testsuite/gcc.target/arc/uncached-stdi-o2.c
+++ b/gcc/testsuite/gcc.target/arc/uncached-stdi-o2.c
@@ -1,0 +1,14 @@
+/* { dg-do compile } */
+/* { dg-options "-O2 -mno-ll64" } */
+
+#include "uncached-stdi.h"
+
+/* First check that all single register uncached store instructions are being generated in the output.  */
+/* { dg-final { scan-assembler "stb.*\\.di" } } */
+/* { dg-final { scan-assembler "\\m(sth|stw).*\\.di" } } */
+/* { dg-final { scan-assembler "st.*\\.di" } } */
+
+/* Second verify that no direct(di) register offset addressing are being generated. */
+/* { dg-final { scan-assembler-not {stb.*\.di\s+r[0-9]+,\s*\[r[0-9]+,\s*r[0-9]+\]} } } */
+/* { dg-final { scan-assembler-not {\m(sth|stw).*\.di\s+r[0-9]+,\s*\[r[0-9]+,\s*r[0-9]+\]} } } */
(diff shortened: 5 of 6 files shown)


gcc/config/arc/arc.md
MichielDerhaeg: Do we also need to fix this?

gcc/testsuite/gcc.target/arc/uncached-stdi.h
```diff
@@ -0,0 +1,24 @@
...
+#define __uncached __attribute__((uncached))
+#endif
+
+volatile char *
+_hl_pack_ptr(volatile char *p, const void *s, int len)
+{
+    volatile __uncached char *buf = p + 4;
+
+    for (int i = 0; i < len; i++)
+        buf[i] = ((const char *)s)[i];
```
MichielDerhaeg, line 11: Should we store more types than `char` to test different load insns?

gcc/config/arc/predicates.md
```diff
@@ -174,6 +174,28 @@
...
+;; memory indexing. Remark that for loads (ld.di) there is support for
+;; register indexing. The predicate blocks the combine optimization
+;; pass from fusing pointer additions back into direct stores preventing
+;; in this way illegal hardware encodings.
+(define_predicate "arc_stdi_memory_operand"
+  (match_code "mem")
+{
+  rtx addr = XEXP (op, 0);
+
+  if (TARGET_V2
```
MichielDerhaeg: Why `TARGET_V2`? Doesn't this just apply to the entire arc target?

gcc/config/arc/arc.md
```diff
@@ -4954,10 +4954,13 @@ archs4x, archs4xd"
...
   [(set_attr "type" "load")])
 
+; Direct store pattern utilizing arc_stdi_memory_operand to
+; reject [reg+reg] on ARCv2. This prevents the combiner from
+; optimizing separate pointer additions back into illegal encodings.
 (define_insn "stdi<mode>"
-  [(unspec_volatile [(match_operand:ALLI 0 "memory_operand"    "m,m,Usc")
-		     (match_operand:ALLI 1 "nonmemory_operand" "r,Cm3,i")]
-		    VUNSPEC_ARC_STDI)]
+  [(unspec_volatile [(match_operand:ALLI 0 "arc_stdi_memory_operand" "m,m,Usc")
```
MichielDerhaeg: Do regular loads also have this issue? If not, why not? How does `*movsi_insn` prevent this from happening?
Perhaps there are other expressions we should also be preventing.

gcc/config/arc/arc.md
```diff
@@ -4954,10 +4954,13 @@ archs4x, archs4xd"
   "ld<mALLI>%U1.di\\t%0,%1"
   [(set_attr "type" "load")])
 
+; Direct store pattern utilizing arc_stdi_memory_operand to
+; reject [reg+reg] on ARCv2. This prevents the combiner from
+; optimizing separate pointer additions back into illegal encodings.
```
MichielDerhaeg: I would not include such verbose comments that just describe the issue you just solved.
It is a clear give-away of AI usage and they won't like it upstream.
I don't think this comment is needed here.

gcc/config/arc/predicates.md
```diff
@@ -174,6 +174,28 @@
   }
 )
 
+;; This predicate is made for direct uncached stores (like ld.di).
+;; It prevents ARCv2 (TARGET_V2) direct stores from using [reg+reg]
+;; memory indexing. Remark that for loads (ld.di) there is support for
+;; register indexing. The predicate blocks the combine optimization
+;; pass from fusing pointer additions back into direct stores preventing
+;; in this way illegal hardware encodings.
```
MichielDerhaeg: Idem, I would keep it less verbose. 1 simple sentence here.

MichielDerhaeg (changes requested):
Do you know which change is responsible for this breaking? Could be useful to know.

This fix will immediately be sent upstream. Can you edit the commit message to just describe the issue you're solving. Something suitable for upstream, we won't reference GH links there.

gcc/config/arc/arc.md
Alexehv77: Ok

gcc/config/arc/arc.md
Alexehv77: Do regular loads also have this issue? 
Regular loads dont have this issues.

If not, why not? 
Only st.di dont accept register memory index operand.
Those stores write directly to the memory - i guess they have to cross longer wires and due to this, timing wise there may not be able to read 3 registers and perform and addition. So you have to bypass the registers (wich involves register address comparators in the bypass network) then perform the address calculation and then send the result to the memory which is further than the cache. 

How does *movsi_insn prevent this from happening?
The movesi_insn doesnt need to prevent this as doesnt cover .di instructions.

Perhaps there are other expressions we should also be preventing.
Im not aware of other expressions.

gcc/config/arc/predicates.md
Alexehv77: ok

gcc/config/arc/predicates.md
Alexehv77: i removed the TARGET_V2 as all arc cores dont support direct stores with register index.

gcc/testsuite/gcc.target/arc/uncached-stdi.h
Alexehv77: we need to test stores but all .di stores for various modes come from the same pattern. I will added anyhow more of them (see the updated file).

Alexehv77 (commented):
I have checked other arc targets and to all of them the direct stores dont support index registers.  From the documentation that i got access i cant say why the hardware doesnt accept such instruction variant. 

I assume that the reason of having this discrepancy can be:
- sw issue regarding the encoding - maybe this kind of addressing is not so frequent so  not supporting it them gives opcode space to be used by other instructions.
- hw issue: to bypass the cache and store directly to the memory requires longer wires and due to this, timing wise there may not be able to read 3 registers and perform and addition.  So you have to bypass the registers (which involves register address comparators in the bypass network) then perform the address calculation and then send the result to the memory which is further than the cache. 

Implementation detail:
During the expand i prevent uncached stores from having register index.
But this is not enough as the combiner can still generate register index addressing. So i still had to make use of a predicate that prevents such addressing. For this I no longer made use of a dedicated predicate and instead i made use of an existent one.

gcc/testsuite/gcc.target/arc/uncached-stdi.h
```diff
@@ -0,0 +1,48 @@
...
+volatile short *
+pack_halfword_ptr(volatile short *p, const void *s, int len)
+{
+    volatile __uncached short *buf = p + 4;
+    for (int i = 0; i < len; i++)
+        buf[i] = ((const short *)s)[i];
+    return p;
+}
+
+/*Test for SImode.  */
```
MichielDerhaeg, line 25: I would also test DImode. Especially because it is a different pattern.

gcc/config/arc/arc.md
MichielDerhaeg: Not handled?

gcc/config/arc/arc.md
MichielDerhaeg: Not handled?

MichielDerhaeg (changes requested):

gcc/config/arc/arc.md
Alexehv77: i removed the comment

gcc/testsuite/gcc.target/arc/uncached-stdi.h
Alexehv77: i added

Alexehv77 (commented):
i added test for the DI mode
there is no need to change that define_insn_and_split

MichielDerhaeg:
```
during RTL pass: split2
uncached-stdi.c: In function 'pack_doubleword_ptr':
uncached-stdi.c:16:1: internal compiler error: in change_address_1, at emit-rtl.cc:2314
   16 | }
      | ^
0x29f4529 diagnostics::context::diagnostic_impl(rich_location*, diagnostics::metadata const*, diagnostics::option_id, char const*, __va_list_tag (*) [1], diagnostics::kind)
        ???:0
0x29eb521 internal_error(char const*, ...)
        ???:0
0x29f493f fancy_abort(char const*, int, char const*)
        ???:0
0x111fcb8 change_address_1(rtx_def*, machine_mode, rtx_def*, int, bool)
        ???:0
0x11202f3 adjust_address_1(rtx_def*, machine_mode, poly_int<1u, long>, int, int, int, poly_int<1u, long>)
        ???:0
0x111e306 gen_highpart(machine_mode, rtx_def*)
        ???:0
0x25ea3a1 gen_split_37(rtx_insn*, rtx_def**)
        ???:0
0x26078a5 split_insns(rtx_def*, rtx_insn*)
        ???:0
0x1123196 try_split(rtx_def*, rtx_insn*, int)
        ???:0
0x15dbfc0 split_insn(rtx_insn*)
        ???:0
0x15dc2dd split_all_insns()
        ???:0
0x15ddfba (anonymous namespace)::pass_split_after_reload::execute(function*)
        ???:0
```

MichielDerhaeg:
> ```
> during RTL pass: split2
> uncached-stdi.c: In function 'pack_doubleword_ptr':
> uncached-stdi.c:16:1: internal compiler error: in change_address_1, at emit-rtl.cc:2314
>    16 | }
>       | ^
> 0x29f4529 diagnostics::context::diagnostic_impl(rich_location*, diagnostics::metadata const*, diagnostics::option_id, char const*, __va_list_tag (*) [1], diagnostics::kind)
>         ???:0
> 0x29eb521 internal_error(char const*, ...)
>         ???:0
> 0x29f493f fancy_abort(char const*, int, char const*)
>         ???:0
> 0x111fcb8 change_address_1(rtx_def*, machine_mode, rtx_def*, int, bool)
>         ???:0
> 0x11202f3 adjust_address_1(rtx_def*, machine_mode, poly_int<1u, long>, int, int, int, poly_int<1u, long>)
>         ???:0
> 0x111e306 gen_highpart(machine_mode, rtx_def*)
>         ???:0
> 0x25ea3a1 gen_split_37(rtx_insn*, rtx_def**)
>         ???:0
> 0x26078a5 split_insns(rtx_def*, rtx_insn*)
>         ???:0
> 0x1123196 try_split(rtx_def*, rtx_insn*, int)
>         ???:0
> 0x15dbfc0 split_insn(rtx_insn*)
>         ???:0
> 0x15dc2dd split_all_insns()
>         ???:0
> 0x15ddfba (anonymous namespace)::pass_split_after_reload::execute(function*)
>         ???:0
> ```

Sorry about the lack of context. I tried building the test with hs38_linux and I mistakenly used -O2 instead of -Os.
-Os works fine, but -O2 crashes..

gcc/config/arc/arc.md
```diff
@@ -4976,10 +4977,22 @@ archs4x, archs4xd"
...
    operands[5] = gen_highpart_mode (SImode, DImode, operands[1]);
-   operands[2] = gen_lowpart (SImode, operands[0]);
-   operands[4] = gen_highpart (SImode, operands[0]);
+
+   rtx addr = XEXP (operands[0], 0);
+   if (GET_CODE (addr) == POST_INC || GET_CODE (addr) == PRE_INC
+       || GET_CODE (addr) == POST_DEC || GET_CODE (addr) == PRE_DEC){
+        rtx base_reg = XEXP (addr, 0);
+        operands[2] = change_address (operands[0], SImode, base_reg);
+        operands[4] = change_address (operands[0], SImode, plus_constant (Pmode, base_reg, 4));
```
MichielDerhaeg: This does not preserve the ´pre/post_inc/dec´. 
Also, in that case, the constant would be used as the inc/dec amount and you can't use ´plus_constant´ to offset the address anymore.
If the post_inc would be preserved, it would increment 4 bytes, instead of the 8 we would want here.
The resulting code is incorrect.

I think it would be best to simply disallow the mem operand to be ´pre/post_inc/dec´. That's the easiest, and definitely correct.

gcc/config/arc/arc.md
```diff
@@ -4976,10 +4977,22 @@ archs4x, archs4xd"
...
+      }
+    else
+      {
+        operands[2] = gen_lowpart (SImode, operands[0]);
+        operands[4] = gen_highpart (SImode, operands[0]);
+      }
   }
   "
+  [(set_attr "type" "store")]
```
MichielDerhaeg: Looks correct, just wondering why you added this.
If there is no good reason I would not pollute the PR with this.

gcc/config/arc/arc.md
```diff
@@ -4944,7 +4944,7 @@ archs4x, archs4xd"
    (set_attr "type" "sr,sr,sr,sr")])
 
 (define_mode_iterator ALLI [QI HI SI (DI "TARGET_LL64")])
-(define_mode_attr mALLI [(QI "b") (HI "%_") (SI "") (DI "d")])
+(define_mode_attr mALLI [(QI "b") (HI "h") (SI "") (DI "d")])
```
MichielDerhaeg: This seemed to be intentional, check arc.cc:5361
´´´
    case '_':
      if (TARGET_V2)
	fputs ("h", file);
      else
	fputs ("w", file);
´´´

gcc/config/arc/arc.md
MichielDerhaeg: What makes you say no changes are needed? I think it should be move_dest_operand or address_operand. I would change it just to be safe.

MichielDerhaeg (changes requested):

gcc/config/arc/arc.md
Alexehv77: Indeed rather than allowing such constructs and then fixing them them i just disabled post/pre inc/dec for DI mode when the target does not have 64 ls unit.

gcc/config/arc/arc.md
Alexehv77: indeed i saw that depending on the core there are both available letters possible

gcc/config/arc/arc.md
Alexehv77: i added it becasue the compiler was crashing without it.

gcc/config/arc/arc.md
Alexehv77: I changed it.
My claim was that the predicate on the define_insn is the one that gonna enforce the generated assembly.
I have changed also that O2 issues which was due to post/pre inc/dec memory constructs - now the predicate takes into account that that post/pre inc/dec should not be allowed for DI mode when the target doesnt support

luismgsilva (approved):

luismgsilva:
Fix your signoff before mergeing.

PR #275, opened 2026-06-27
arc64: Fix for P10019563-86826

Alexehv77:
**arc64: Fix crash in emit_move_multi_word during rtl expansion**

Fix for multiword moves on hs5x introduced during:
define_expand "<ANY_EXTEND:su_optab>maddsidi4" 
Originally when this expand was trying to move data in DI mode
to the reg 58 the gcc internal emit_move_insn function had inside an issue: 
Inside  there is the following loop which endsup trying to make use of reg 59 which is reserved.

**Namely:**
```
for (i = 0; i < 2; i++) {
rtx xpart = operand_subword (x, i, 1, mode);
:
   gcc_assert (xpart && ypart); //at the second run (i=1) tries
                                //to take r59 and it crashes.
:
}
```
**Solution:**
Remove the use of emit_move_insn and pass the operands directly into the gen_<ANY_EXTEND:su_optab>macd bypassing
the logic inside emit_move_insn. The <ANY_EXTEND:su_optab>macd accepts r58 and produces a pattern that is subject to a define_insn_and_split.

(define_insn_and_split "<ANY_EXTEND:su_optab>macd"
This split takes place after ra and already contains the move (set (reg:DI R58_REGNUM) (match_dup 3) generating
a move (vadd2) followed by a macdu. For more info check:
https://pyreneesgf.atlassian.net/browse/P10019563-86826

### DIFF
--- a/gcc/config/arc64/arith.md
+++ b/gcc/config/arc64/arith.md
@@ -1062,20 +1062,17 @@
 ;; 64 + (signe) 32 x (signe) 32 -> 64
 (define_expand "<ANY_EXTEND:su_optab>maddsidi4"
   [(set (match_operand: DI 0 "register_operand")
-	(plus:DI
-	 (mult:DI
-	  (ANY_EXTEND:DI (match_operand:SI 1 "register_operand"))
-	  (ANY_EXTEND:DI (match_operand:SI 2 "register_operand")))
-	 (match_operand:DI 3 "register_operand")))]
+        (plus:DI
+         (mult:DI
+          (ANY_EXTEND:DI (match_operand:SI 1 "register_operand"))
+          (ANY_EXTEND:DI (match_operand:SI 2 "register_operand")))
+         (match_operand:DI 3 "register_operand")))]
   "TARGET_SIMD"
   {
-   rtx acc = gen_rtx_REG (DImode, R58_REGNUM);
-
-   emit_move_insn (acc, operands[3]);
    emit_insn (gen_<ANY_EXTEND:su_optab>macd (operands[0], operands[1],
-                                            operands[2], acc));
+                                            operands[2], operands[3]));
    DONE;
-   })
+  })
 
 (define_insn "<ANY_EXTEND:su_optab>macd0"
  [(set (reg:DI R58_REGNUM)
--- a/gcc/testsuite/gcc.target/arc64/maddsidi4-simd.c
+++ b/gcc/testsuite/gcc.target/arc64/maddsidi4-simd.c
@@ -1,0 +1,40 @@
+/* { dg-do compile } */
+/* { dg-options "-mcpu=hs5x -O2" } */
+
+#include <stdint.h>
+
+uint64_t
+__umul128(const uint64_t a, const uint64_t b, uint64_t * const productHi)
+{
+    const uint32_t aLo = (uint32_t)a;
+    const uint32_t aHi = (uint32_t)(a >> 32);
+    const uint32_t bLo = (uint32_t)b;
+    const uint32_t bHi = (uint32_t)(b >> 32);
+
+    const uint64_t b00 = (uint64_t)aLo * bLo;
+    const uint64_t b01 = (uint64_t)aLo * bHi;
+    const uint64_t b10 = (uint64_t)aHi * bLo;
+    const uint64_t b11 = (uint64_t)aHi * bHi;
+
+    const uint32_t b00Lo = (uint32_t)b00;
+    const uint32_t b00Hi = (uint32_t)(b00 >> 32);
+
+    const uint64_t mid1 = b10 + b00Hi;
+    const uint32_t mid1Lo = (uint32_t)(mid1);
+    const uint32_t mid1Hi = (uint32_t)(mid1 >> 32);
+
+    const uint64_t mid2 = b01 + mid1Lo;
+    const uint32_t mid2Lo = (uint32_t)(mid2);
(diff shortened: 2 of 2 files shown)


gcc/config/arc64/arith.md
```diff
@@ -1059,23 +1059,38 @@
...
   "TARGET_SIMD"
   {
-   rtx acc = gen_rtx_REG (DImode, R58_REGNUM);
-
-   emit_move_insn (acc, operands[3]);
    emit_insn (gen_<ANY_EXTEND:su_optab>macd (operands[0], operands[1],
-                                            operands[2], acc));
+                                            operands[2], operands[3]));
    DONE;
-   })
```
MichielDerhaeg, line 1078: Can this be simplified further? Produce the macd pattern directly? Or should macd be renamed to this one?

MichielDerhaeg (changes requested):
This need a description on why the old code is wrong (it isn't obvious to me at least). And what exactly this fixes.

Alexehv77:
> This need a description on why the old code is wrong (it isn't obvious to me at least). And what exactly this fixes.

While trying to move data in DI mode to reg 58 the emit_move_insn had inside an issue.
Inside this function there is this loop which endsup trying to make use of reg 59 which is reserved.

Namely:
for (i = 0; i < 2; i++) {
    rtx xpart = operand_subword (x, i, 1, mode);….
    .……….
    gcc_assert (xpart && ypart); // at the second run (i=1) tries to take r59 and it crashes.
}

Solution:
Remove the use of emit_move_insn and pass the operands directly into the 
gen_<ANY_EXTEND:su_optab>macd
bypassing the logic inside emit_move_insn. 

This <ANY_EXTEND:su_optab>macd accepts r58 and produces a pattern that is subject to a define_insn_and_split. 

(define_insn_and_split "<ANY_EXTEND:su_optab>macd"

This split takes place after ra and already contains  the move (set (reg:DI R58_REGNUM) (match_dup 3) generating a move (vadd2) followed by a macdu.

gcc/config/arc64/arith.md
Alexehv77: There is no define_insn for it but a define_insn_and_split.
As such i could merge it but im afraid not to trigger issuesas the whole point was to avoid letting the compiler being able to reason about r58.

Alexehv77 (commented):
i added extra comments.

MichielDerhaeg:
> > This need a description on why the old code is wrong (it isn't obvious to me at least). And what exactly this fixes.
> 
> While trying to move data in DI mode to reg 58 the emit_move_insn had inside an issue. Inside this function there is this loop which endsup trying to make use of reg 59 which is reserved.
> 
> Namely: for (i = 0; i < 2; i++) { rtx xpart = operand_subword (x, i, 1, mode);…. .………. gcc_assert (xpart && ypart); // at the second run (i=1) tries to take r59 and it crashes. }
> 
> Solution: Remove the use of emit_move_insn and pass the operands directly into the gen_<ANY_EXTEND:su_optab>macd bypassing the logic inside emit_move_insn.
> 
> This <ANY_EXTEND:su_optab>macd accepts r58 and produces a pattern that is subject to a define_insn_and_split.
> 
> (define_insn_and_split "<ANY_EXTEND:su_optab>macd"
> 
> This split takes place after ra and already contains the move (set (reg:DI R58_REGNUM) (match_dup 3) generating a move (vadd2) followed by a macdu.

Can you put (some version of) this in the commit message?

gcc/config/arc64/arith.md
MichielDerhaeg: > As such i could merge it but im afraid not to trigger issuesas the whole point was to avoid letting the compiler being able to reason about r58.

I think its semantically equivalent. You're generating the exact same RTL either way.

MichielDerhaeg:
> > > This need a description on why the old code is wrong (it isn't obvious to me at least). And what exactly this fixes.
> > 
> > 
> > While trying to move data in DI mode to reg 58 the emit_move_insn had inside an issue. Inside this function there is this loop which endsup trying to make use of reg 59 which is reserved.
> > Namely: for (i = 0; i < 2; i++) { rtx xpart = operand_subword (x, i, 1, mode);…. .………. gcc_assert (xpart && ypart); // at the second run (i=1) tries to take r59 and it crashes. }
> > Solution: Remove the use of emit_move_insn and pass the operands directly into the gen_<ANY_EXTEND:su_optab>macd bypassing the logic inside emit_move_insn.
> > This <ANY_EXTEND:su_optab>macd accepts r58 and produces a pattern that is subject to a define_insn_and_split.
> > (define_insn_and_split "<ANY_EXTEND:su_optab>macd"
> > This split takes place after ra and already contains the move (set (reg:DI R58_REGNUM) (match_dup 3) generating a move (vadd2) followed by a macdu.
> 
> Can you put (some version of) this in the commit message?

Is this waiting on me? I think the RTL is fine, whether the expander can be merged with the insn i leave up to you. But I do want the commit message to contain more info.

gcc/config/arc64/arith.md
Alexehv77: the mac cant be combine with the maddsidi4 as the mac requires an extra paramenter. On the other hand we cant get rid of maddsidi4 as is a predefined name known at the expander level. Also the mac is being used inside the arc64.cc file. 
But i think there is space to improve in general the arith.md file in the arc64.md and for this i added a jira ticket with my observations:
https://pyreneesgf.atlassian.net/browse/P10019563-87259

MichielDerhaeg (approved):

PR #277, opened 2026-06-30
arcv: Verify RTL when checking for fusions

MichielDerhaeg:


gcc/config/riscv/arcv.cc
```diff
@@ -338,80 +338,117 @@ arcv_ls_update (rtx_insn *prev, rtx_insn *curr)
...
+	rtx base, offset;
+	if (!extract_base_offset_in_addr (p_mem, &base, &offset)
+	    || !REG_P (p_dest))
+	  return false;
+
+	int p_rs = REGNO (base);
+	int p_rd = REGNO (p_dest);
 
-	return (p_rs == c_rs1
+	return ((p_rs == c_rs1 || p_rs == c_rs2)
```
exur00: p_rs is rs1 (base) of the load, why check agains rs2?

exur00: no performance change, but seems more correct. I'll change it in my branch where I'm also adding the c.mv

gcc/config/riscv/arcv.cc
```diff
@@ -338,80 +338,117 @@ arcv_ls_update (rtx_insn *prev, rtx_insn *curr)
...
+	if (!extract_base_offset_in_addr (p_mem, &base, &offset))
 	  return false;
 
-	int p_rs = REGNO (p_dst_addr);
+	int p_rs = REGNO (base);
 
-	return (p_rs == c_rs1
-		&& (!has_rs2 || /* Fused ST + OP.  */
-		    p_rs == c_rs2)); /* Fused ST + OP-IMM.  */
+	if (p_rs != c_rs1 && p_rs != c_rs2)
```
exur00: same question

gcc/config/riscv/arcv.cc
```diff
@@ -338,80 +338,117 @@ arcv_ls_update (rtx_insn *prev, rtx_insn *curr)
...
+
+	if (c_rs2 == -1)
+	  return true;
+
+	rtx data = SET_SRC (prev_set);
+	if (!REG_P (data))
+	  return false;
+
+	int p_rs2 = REGNO (data);
+	return (p_rs2 == c_rs1 || p_rs2 == c_rs2);
```
exur00: here too

MichielDerhaeg:
Can this get an approval?

exur00 (approved):
looks better and more correct. Performance is good

gcc/config/riscv/arcv.cc
MichielDerhaeg: I changed it back. I interpreted it as the amount of source registers need to be below 2. But now I'm not so sure whether the order matters.

In any case. This PR should handle the correctness of the original intent. We should investigate what the hardware does before relaxing the order of rs1 & rs2.

luismgsilva (approved):

exur00:
I found that 
```
;;	  0--> b  0: i 821 a5=a5+0x4
ARCV_FUSE_LS_UPDATE (822, 821)
;;	  0--> b  0: i 822 a3=[a5] 
```
is still falsely detected as fusion in this patch, I will investigate and fix

exur00:
> I found that
> 
> ```
> ;;	  0--> b  0: i 821 a5=a5+0x4
> ARCV_FUSE_LS_UPDATE (822, 821)
> ;;	  0--> b  0: i 822 a3=[a5] 
> ```
> 
> is still falsely detected as fusion in this patch, I will investigate and fix

checking this does break the current symmetry:
```
d = [b + offset]
a = b + c
```
here d cannot be the same register as b or there is a RAW hazard that is currently not checked.
```
a = b + c
d = [b + offset]
```
has no issue if b and d are the same register.

MichielDerhaeg:
> I found that
> 
> ```
> ;;	  0--> b  0: i 821 a5=a5+0x4
> ARCV_FUSE_LS_UPDATE (822, 821)
> ;;	  0--> b  0: i 822 a3=[a5] 
> ```
> 
> is still falsely detected as fusion in this patch, I will investigate and fix

We opted not to include that in this PR, and continued with the previous implementation so it can merge. Incremental improvements can be made later.

PR #284, opened 2026-07-07
Fix RISCV_FUSE_MULT_ADD matching for RV64 sign-extended ops.

luismgsilva:
```
fixup! RISC-V: Add instruction patterns for 32-bit multiply-add and bit-extract fusion.

tmp: Fix RV64 madd_fused_extended split to use SI lowpart operand.

When madd_fused_extended splits with the multiply result left in the
destination register, the subsequent addsi3_extended must take the
first plus:SI operand from gen_lowpart (SImode, operands[0]).  After
mulsi3_extended, operands[0] is reg:DI, so passing it directly
produces invalid RTL such as:

  (sign_extend:DI (plus:SI (reg:DI ...) (reg:SI ...)))

which no insn pattern matches, causing an ICE during split2 on RV64
RHX-100 builds (e.g. picolibc hash_page.c).

Signed-off-by: Luis Silva <luis.silva@globalfoundries.com>
```

```
fixup! RISC-V: Implement riscv_macro_fusion_pair_p for Synopsys RHX-100 series.
    
tmp: Fix RISCV_FUSE_MULT_ADD matching for RV64 sign-extended ops.
   
On RV64, 32-bit multiply-add uses SI-mode RTL sign-extended to DI
(e.g. mulsi3_extended and addsi3_extended), but riscv_fuse_mult_add
only looked for top-level MULT and PLUS nodes.  This failed to match
fused sequences on RV64.

Signed-off-by: Luis Silva <luis.silva@globalfoundries.com>
```

### DIFF
--- a/gcc/config/riscv/riscv-fusion.cc
+++ b/gcc/config/riscv/riscv-fusion.cc
@@ -932,8 +932,13 @@ riscv_fuse_b_alui (rtx_insn *prev, rtx_insn *curr)
 }
 
 /* Check for RISCV_FUSE_MULT_ADD fusion.
-   prev (mult) == (set (reg:DI rD_mult) (mult:DI (reg:DI rS1) (reg:DI rS2)))
-   curr (add)  == (set (reg:DI rD_add) (plus:DI (reg:DI rD_mult) (reg:DI rS3)))  */
+   On RV32:
+   prev (mul) == (set (reg rD) (mult:SI (reg:SI rS1) (reg:SI rS2)))
+   curr (add) == (set (reg rD) (plus:SI (reg:SI rD) (reg:SI rS3)))
+
+   On RV64 the result may be sign-extended to DI:
+   prev (mul) == (set (reg:DI rD) (sign_extend:DI (mult:SI ...)))
+   curr (add) == (set (reg:DI rD) (sign_extend:DI (plus:SI ...)))  */
 
 static bool
 riscv_fuse_mult_add (rtx_insn *prev, rtx_insn *curr)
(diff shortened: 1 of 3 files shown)


gcc/config/riscv/riscv.md
```diff
@@ -4802,7 +4802,9 @@
      else
        {
 	 emit_insn (gen_mulsi3_extended (operands[0], operands[1], operands[2]));
-	 emit_insn (gen_addsi3_extended (operands[0], operands[0], operands[3]));
+	 emit_insn (gen_addsi3_extended (operands[0],
+					 gen_lowpart (SImode, operands[0]),
```
MichielDerhaeg, line 4807: Why is this suddenly needed?

luismgsilva: If you build rv64 with rhx mtune it crashes

EDIT: Even on older version of `michiel/fusion-trunk-3`

gcc/testsuite/gcc.target/riscv/arcv-fusion-mult-add-rv64-dump.c
```diff
@@ -0,0 +1,12 @@
+/* { dg-do compile } */
+/* { dg-skip-if "" { *-*-* } { "-g" "-flto" "-O0" "-O1" "-O3" "-Oz" "-Os" } } */
+/* { dg-options "-O2 -mtune=arc-v-rhx-100-series -march=rv64im -mabi=lp64 -fdump-rtl-sched2" { target { rv64 } } } */
+
+int
+fuse_mult_add_rv64 (int a, int b, int c)
+{
+  int m = a * b;
```
MichielDerhaeg: Shouldn't there be a test for this already for rv32? If so, let's just merge them

gcc/testsuite/gcc.target/riscv/arcv-fusion-mult-add-rv64-dump.c
```diff
@@ -0,0 +1,12 @@
...
+  return m + c;
+}
+
+/* { dg-final { scan-rtl-dump "RISCV_FUSE_MULT_ADD" "sched2" { target { rv64 } } } } */
```
MichielDerhaeg: No reason for this to have a target requirement if merged.

gcc/config/riscv/riscv-fusion.cc
```diff
@@ -932,8 +932,8 @@ riscv_fuse_b_alui (rtx_insn *prev, rtx_insn *curr)
 }
 
 /* Check for RISCV_FUSE_MULT_ADD fusion.
-   prev (mult) == (set (reg:DI rD_mult) (mult:DI (reg:DI rS1) (reg:DI rS2)))
-   curr (add)  == (set (reg:DI rD_add) (plus:DI (reg:DI rD_mult) (reg:DI rS3)))  */
+   prev (mul) == (set (reg rD) (mult:SI (reg:SI rS1) (reg:SI rS2)))
+   curr (add) == (set (reg rD) (plus:SI (reg:SI rD) (reg:SI rS3)))  */
```
MichielDerhaeg: We should document the sign-extended variant as well.

gcc/config/riscv/riscv.md
MichielDerhaeg: This also needs to be fixed for `operands[4]´

gcc/testsuite/gcc.target/riscv/arcv-fusion-mult-add-rv64-dump.c
luismgsilva: Done. I merged it into the existing op0 test and added the RV64 dg-options. The test now returns `m + c` because `m + c + m + d` widens to DI on RV64 and doesn't fuse. The `op1` test stays RV32 because on RV64 it ends up testing `op0`, so it isn't really checking `op1`.

gcc/config/riscv/riscv-fusion.cc
luismgsilva: See https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/284/changes/dad04330f0b15d2932354483e606b68940792655

gcc/config/riscv/riscv.md
luismgsilva: See https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/284/changes/03e4f18bee3a26243244c82c45b99642a943a55a

MichielDerhaeg (approved):

PR #285, opened 2026-07-08
arcv: fix vsetvl error due to wrong insn pattern

Alexehv77:
The compiler aborted inside the vsetvl pass during has_vl(): gcc_assert (!m_vl || REG_P (m_vl))

The attribute vl_op_idx was set to 4 as if was an instruction with immediate payload. 
Had to be changed to index 5 which holds the vector length operand.
Also the avl_type_idx was set to 7 while it had to be 8.
Also a dejagnu test needed to be changed - but thats orthogonal to this issue in sense that was wrong test anyhow.
Extra info can be found here:
https://pyreneesgf.atlassian.net/browse/P10019563-86957

### DIFF
--- a/gcc/config/riscv/arcv-vector.md
+++ b/gcc/config/riscv/arcv-vector.md
@@ -281,10 +281,10 @@
   "arcv.vclr.v.i\t%0,%4%p1"
   [(set_attr "type" "viwmuladd")
    (set_attr "mode" "<MODE>")
-   (set_attr "vl_op_idx" "4")
+   (set_attr "vl_op_idx" "5")
    (set (attr "ta") (symbol_ref "riscv_vector::get_ta(operands[5])"))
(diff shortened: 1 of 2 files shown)


luismgsilva:
Just FYI, this instruction has changed before, but the patch was never merged due to internal architecture discussions. See PR: https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/pull/206
There are also related discussions here: [P10019563-62986](https://pyreneesgf.atlassian.net/browse/P10019563-62986)
It will change again as part of: [P10019563-86588](https://pyreneesgf.atlassian.net/browse/P10019563-86588)

Either way, this looks good to me. The attributes were also something I didn't address in my initial patch.

luismgsilva (approved):

MichielDerhaeg (approved):

PR #287, opened 2026-07-09
tmp: add option to disable for testing

exur00:
rmx500 does not have adjacent load-store fusions.


exur00:
TODO:
- [ ] proper commit message
- [ ] set off by default for rmx500
- [ ] performance:
  - [ ]  4f83c189f4c9cc83b8e4ac63614a5c5b27f58585 and/or 423a4122f5ce45d89e3fdb1d7b9e1820e8a5cd83 combined with this PR causes 5% coremark regression, both separately not
  - [ ] embench huffbench 4.3% regression, I expect to be solved by !277
  - [ ] embench sglib and slre 0.6% regressions, I also expect to be solved by !277

PR #288, opened 2026-07-09
arcv: check for c.mv for load/store-op fusions

exur00:
Contrary to mv, c.mv rd, rs expands such that rs is the 2nd source register. rs1 is 0, which is exempt for the existing check. Added that exemption as well.

Signed-off-by: Loeka Rogge loeka.rogge@gf.com


exur00:
TODO:
- [ ] Add test
  - [x] c.mv
  - [ ] x0 exception
- [ ] Fix errors
- [x] now we discovered this exception, re-add the li/addi ls_update, as was in the !276 draft
- [x] check if x0 exception also applies for (I1.rs2 == I2.rs2) for store-op fusion, if so, add it

exur00:
checking for GP_REG_FIRST (x0) makes no sense, I think it's still const_int 0 here instead of register x0

gcc/config/riscv/arcv.cc
```diff
@@ -321,6 +321,27 @@ arcv_fusible_memop_p (rtx_insn *insn)
 }
 
 
+static bool
+compressed_move (rtx_insn *insn)
+  {
+    if (!TARGET_ZCA)
+      return false;
```
MichielDerhaeg: The compressed extension has been split up. You need to check both `TARGET_RVC` and `TARGET_ZCA`

gcc/config/riscv/arcv.cc
```diff
@@ -321,6 +321,27 @@ arcv_fusible_memop_p (rtx_insn *insn)
...
+
+    gcc_assert (REG_P (SET_SRC (insn_set)));
+    int rs = REGNO (SET_SRC (insn_set));
+
+    gcc_assert (REG_P (SET_DEST (insn_set)));
+    int rd = REGNO (SET_DEST (insn_set));
+
+    return riscv_compressed_reg_p (rs) && riscv_compressed_reg_p (rd);
+  }
+
```
MichielDerhaeg: I think you can make this really small if you inline it and put it under the `REG_P (c_csr)`? The target is already checked in `riscv_compressed_reg_p` and `TYPE_MOVE` would need to be check it should've already been done?

exur00:
existing logic checks both rs1 and rs2 of the OP against the rs1 of a store, instead of rs1 and rs2 respectively

exur00:
performance impact on rhx negligible (smallest change 0.06%, some worse, some better)
standalone impact on rmx500 quite small, but combined with !287 improves still a bit above both separately (additinal 0.45% on coremark, a few similar embench improvements)

exur00:
> existing logic checks both rs1 and rs2 of the OP against the rs1 of a store, instead of rs1 and rs2 respectively

fixed

MichielDerhaeg:
Note to self, we need to communicate these findings to the MW team.

exur00:
The 2 simulators disagree on this case, we should probably wait to apply this until we have clarity

PR #289, opened 2026-07-09
tmp: Remove TARGET_ARCV_FUSION and TARGET_ARCV_RHX100

luismgsilva:
in favor of checking for specific fusion.


gcc/config/riscv/riscv.cc
```diff
@@ -11279,15 +11280,15 @@ riscv_sched_init (FILE *, int, int)
 {
   clear_vconfig ();
 
-  if (TARGET_ARCV_FUSION)
+  if (riscv_macro_fusion_p ())
     arcv_sched_init ();
```
MichielDerhaeg: All of these that call into `arc_*` should be `TARGET_ARCV_RHX100`

luismgsilva: Done

MichielDerhaeg (approved):

PR #290, opened 2026-07-09
arcv: Fix arcv_rhx100 tuning costs causing unexpected spill regressions

Alexehv77:
While running the xtheadfmv-fmv.c test case for -march=rv32gc_xtheadfmv, the compiler was dropping the direct register move instructions (th.fmv.hw.x). Instead it fell back to spilling 64 bit data onto the stack using store (sw) and load (fld) loops.

This issue occurred because our custom tuning profile configurations (--with-tune=arc-v-rhx-100-series and also for rmx100, rmx500, and rpx100 cores) had an internal cost model which gave priority to load/store rather than moves.

Correcting the core weights eliminated the slow memory operations and as a result resolved 40 C test file failures (amounting to a few hundred total tests when accounting for optimization flags).

jira ticket:
https://pyreneesgf.atlassian.net/browse/P10019563-86975

### DIFF
--- a/gcc/config/riscv/riscv.cc
+++ b/gcc/config/riscv/riscv.cc
@@ -850,7 +850,7 @@ static const struct riscv_tune_param arcv_rmx100_tune_info = {
   {COSTS_N_INSNS (17), COSTS_N_INSNS (17)},	/* int_div */
   1,						/* issue_rate */
   4,						/* branch_cost */
-  2,						/* memory_cost */
+  4,						/* memory_cost */
   4,						/* fmv_cost */
   false,					/* slow_unaligned_access */
   false,					/* vector_unaligned_access */
@@ -875,7 +875,7 @@ static const struct riscv_tune_param arcv_rmx500_tune_info = {
   {COSTS_N_INSNS (21), COSTS_N_INSNS (21)},	/* int_div */
   2,						/* issue_rate */
   9,						/* branch_cost */
-  2,						/* memory_cost */
+  4,						/* memory_cost */
   8,						/* fmv_cost */
   false,					/* slow_unaligned_access */
   false,					/* vector_unaligned_access */
@@ -900,7 +900,7 @@ static const struct riscv_tune_param arcv_rhx100_tune_info = {
   {COSTS_N_INSNS (27), COSTS_N_INSNS (43)},	/* int_div */
   4,						/* issue_rate */
   9,						/* branch_cost */
-  2,						/* memory_cost */
+  4,						/* memory_cost */
   8,						/* fmv_cost */
   false,					/* slow_unaligned_access */
   false,					/* vector_unaligned_access */
@@ -925,7 +925,7 @@ static const struct riscv_tune_param arcv_rpx100_tune_info = {
   {COSTS_N_INSNS (27), COSTS_N_INSNS (43)},	/* int_div */
   4,						/* issue_rate */
   9,						/* branch_cost */
-  2,						/* memory_cost */
+  4,						/* memory_cost */
   8,						/* fmv_cost */
   false,					/* slow_unaligned_access */
   false,					/* vector_unaligned_access */


MichielDerhaeg:
If tests assume a higher memory cost, that doesn't mean ours is wrong.
This needs to be benchmarked first.

Fairly certain this is not correct for rmx100, when data is in iccm, loads and stores take 1 cycle, moving through memory is cheap there. With caching and store graduation, same can be true of the other cores.

Alexehv77:
I did some analysis of this patch. Please have a look to my explanations.

First of all the costs that we use at least for some of the tunings like rhx100 are not proper.
For example giving for memory costs (memory_cost) a value of 2 for all the core variants is very low 
which as result will encourage ra to use memory slots (almost) freely.

Remark also that there is no other riscv variant provided by any other vendor
with so low value. If you setup the memory_cost to 2 then this is the same as the cost of
the gp to gp registers. 

See here the value 2 returned by this hook:

```
int riscv_register_move_cost (machine_mode mode,
                          reg_class_t from, reg_class_t to)
{
:
:
  return riscv_secondary_memory_needed (mode, from, to) ? 8 : 2;
}

```
Further fmv_cost represents the cost assigned to moving data between 
disjoint register files like general and fp. Setting up the value of 8
will discourage moving data between them but this should not prohibit 
direct moves if they exist.
Now ratio wise 8/2 leads to inefficient code. Here is such an example but you can find 
about 40 more such example in dejagnu 

For example when compiling this code:
```

double ll2d (long long ll)
{
  return *(double*)&ll;
}

```
Before ra there is this code sequence:
```

(insn 13 4 2 2 (set (reg:DI 137 [ ll ])
        (reg:DI 10 a0 [ ll ])) "xtheadfmv-fmv.c":7:1 277 {*movdi_32bit}
     (expr_list:REG_DEAD (reg:DI 10 a0 [ ll ])
        (nil)))
(note 2 13 3 2 NOTE_INSN_DELETED)
(note 3 2 10 2 NOTE_INSN_FUNCTION_BEG)
(insn 10 3 11 2 (set (reg/i:DF 42 fa0)
        (subreg:DF (reg:DI 137 [ ll ]) 0)) "xtheadfmv-fmv.c":9:1 288 {*movdf_hardfloat_rv32}
     (expr_list:REG_DEAD (reg:DI 137 [ ll ])
        (nil)))
(insn 11 10 0 2 (use (reg/i:DF 42 fa0)) "xtheadfmv-fmv.c":9:1 -1
     (nil))

```
gcc16 picks instructions following inside each ebb the order from top to bottom.
It picks first insn 13 (during reload an unallocated pseudo goes to stack - unless there 
is some further understanding
of its use - but here thats not the case see in the intermediary code this message:
"Rejecting inheritance for 137 because of disjoint classes" ) 
and based on the following analysis ends up producing a store instruction.

As you see it looks also to whom consumes the output of insn 13 - namely insn 10.
```

      Considering alt=0 of insn 13:   (0) =r  (1) r
            0 Operand reload: losers++
            0 Non input pseudo reload: reject++
            Cycle danger: overall += LRA_MAX_REJECT
          overall=607,losers=1,rld_nregs=2
         Considering alt=1 of insn 13:   (0) r  (1) i
            0 Operand reload: losers++
            0 Non input pseudo reload: reject++
            Bad operand -- refuse
         Considering alt=2 of insn 13:   (0) r  (1) m
            0 Operand reload: losers++
            0 Non input pseudo reload: reject++
            1 Operand reload: losers++
            1 Non-pseudo reload: reject+=2
            1 Spill pseudo into memory: reject+=3
            Using memory insn operand 1: reject+=3
            Cycle danger: overall += LRA_MAX_REJECT
          overall=621,losers=2,rld_nregs=2
         Considering alt=3 of insn 13:   (0) m  (1) rJ
            0 Non pseudo reload: reject++
          overall=1,losers=0,rld_nregs=0
      Choosing alt 3 in insn 13:  (0) m  (1) rJ {*movdi_32bit}
         Considering alt=0 of insn 10:   (0) =f  (1) f
            1 Operand reload: losers++
            Cycle danger: overall += LRA_MAX_REJECT
          overall=606,losers=1,rld_nregs=1
         Considering alt=1 of insn 10:   (0) f  (1) zfli
            Bad operand -- refuse
         Considering alt=2 of insn 10:   (0) f  (1) G
            Bad operand -- refuse
         Considering alt=3 of insn 10:   (0) f  (1) m
            1 Non pseudo reload: reject++
          overall=1,losers=0,rld_nregs=0
      Choosing alt 3 in insn 10:  (0) f  (1) m {*movdf_hardfloat_rv32}
           Spilling non-eliminable hard regs: 2
```


Bacause the inheritance to insn 10 consumption fails then
insn 13 is treated in isolation and there the memory cost
is low which results in creating the store variant of movdi_32bit.
The inheritance fails because the output of insn13 is 64 bit and
needs 2 gp registers. But the consumption is by a subreg:DF of the output of insn 13
and this output seats on a single fp register. Not knowing 
how to deal from the 2 gp regs to the 1 fp regs 
(see the msg "Rejecting inheritance for 137 because of disjoint classes") 
causes to treat insn 13 in isolation and to choose for it to be recognized 
as a store to the stack.

Further once the choice for insn13 is made the choice for insn10 comes implicitly.
For insn 10 its input is no longer an unallocated register but rather a memory location.
```

         Spilling non-eliminable hard regs: 2
         Considering alt=0 of insn 10:   (0) =f  (1) f
            1 Operand reload: losers++
            1 Non-pseudo reload: reject+=2
            1 Non input pseudo reload: reject++
            Cycle danger: overall += LRA_MAX_REJECT
          overall=609,losers=1,rld_nregs=1
         Considering alt=1 of insn 10:   (0) f  (1) zfli
            Bad operand -- refuse
         Considering alt=2 of insn 10:   (0) f  (1) G
            Bad operand -- refuse
         Considering alt=3 of insn 10:   (0) f  (1) m
          overall=0,losers=0,rld_nregs=0
      Choosing alt 3 in insn 10:  (0) f  (1) m {*movdf_hardfloat_rv32}
         Considering alt=3 of insn 13:   (0) m  (1) rJ
          overall=0,losers=0,rld_nregs=0
      Choosing alt 3 in insn 13:  (0) m  (1) rJ {*movdi_32bit}
           Spilling non-eliminable hard regs: 2
```

To recap the low memory cost coupled with not being able to connect the consumption of
gpr by a fp register ends up producing the following code:

        addi    sp,sp,-16
        sw      a0,8(sp)
        sw      a1,12(sp)
        fld     fa0,8(sp)
        addi    sp,sp,16

Alternatively - similar to llvm and gcc15 -  we could have generated this code:

    fmv.w.x fa0,a0
    th.fmv.hw.x     fa0,a1

Regarding gcc 15 - although there were the same tuning costs i observed
that the reload starts from bottom of the block. Here it inheritance works.
Base on it the compiler calculates that is cheaper to use registers than
memory despite high cost of registers. The main reason behind this is 
using registers implies that there are no emergency spills needed so the cost is 8.
There is further complexity regarding the choice of going to memory which falls out 
of the scope of this description.
The point is that the comparison between the 2 choices i was able to play with it.

There are several solutions to address this issue:

- Change the costs of multiword move costs inside riscv_memory_move_cost

This is what my current patch does. 
Is simple yet effective.
In this phase of the work it doesn't make sense to address performance 
tunning while there are still dejagnu failures.

- Change riscv_secondary_memory_needed when dealing with 64 bit data moves 

between gp registers and fp registers.
Currently this hook is implemented such that when floating point registers 
are wider than integer ones moves between them must go through memory.
I can change this to make the compiler aware that there are direct moves so 
no memory needed. As a result the reload will be able to take into account
that there are direct moves without even bothering checking for memory 
passing alternatives.

- Fix the inheritance issue observed from insn13 to insn 10.

This can be done implementing riscv_secondary_reload. We can look to the 
arc_secondary_reload. Here we need to be careful because TARGET_SECONDARY_RELOAD is 
not needed by any other target so far.

I prefer you accept my current changes taking into account that 
refinements or alternative solutions can be considered later when
we will address performance.

MichielDerhaeg:
> Now ratio wise 8/2 leads to inefficient code. Here is such an example but you can find
> about 40 more such example in dejagnu

The fmv costs is probably also just way too high. Something for @exur00 to look at perhaps when he starts evaluating floating point performance.

> In this phase of the work it doesn't make sense to address performance
> tunning while there are still dejagnu failures.

But you're changing costs, how is this not performance tuning?

> I prefer you accept my current changes taking into account that
> refinements or alternative solutions can be considered later when
> we will address performance.

I strongly suggest you wait with changing the costs until you start systematically addressing performance.
Otherwise this will just add noise to the benchmark results. It'll need to be reverted to establish a correct baseline.

I strongly believe you are correct that the current costs are wrong. But I'd really prefer not to change them until we have measured the effects. That dejagnu tests are currently failing because of this is not a problem, these tests can be ignored for now.

Let's come back to this when you've addressed all *functional* issues.

PR #298, opened 2026-07-17
RHX100 Fusion merge lui_ld

luismgsilva:


gcc/config/riscv/riscv.cc
```diff
@@ -984,7 +984,7 @@ static const struct riscv_tune_param arcv_rhx100_tune_info = {
    | RISCV_FUSE_ADJACENT_LOAD | RISCV_FUSE_ADJACENT_STORE
    | RISCV_FUSE_LS_UPDATE | RISCV_FUSE_LUI_ST
    | RISCV_FUSE_LI_STORE | RISCV_FUSE_BFEXT
-   | RISCV_FUSE_LUI_LD | RISCV_FUSE_LUI_LD_REV),  /* fusible_ops */
+   | RISCV_FUSE_LUI_LD_DEST_DIFF),  /* fusible_ops */
```
MichielDerhaeg: Make sure the comment stays aligned

gcc/config/riscv/riscv-fusion.cc
```diff
@@ -1160,30 +1160,41 @@ riscv_fuse_li_store (rtx_insn *prev, rtx_insn *curr)
...
+	  && REG_P (SET_DEST (lui_set))
+	  && ((get_attr_type (lui) == TYPE_MOVE
+	       && GET_CODE (SET_SRC (lui_set)) == HIGH)
+	      || (CONST_INT_P (SET_SRC (lui_set))
+		  && LUI_NONZERO_OPERAND (INTVAL (SET_SRC (lui_set)))))
+	  && REGNO (SET_DEST (load_set)) != REGNO (SET_DEST (lui_set)));
+}
+
+/* Check for RISCV_FUSE_LUI_LD_DEST_DIFF fusion.
    prev (ld)  == (set (reg:DI rD1)
```
MichielDerhaeg: I'm very sorry that I keep changing my mind on this, but I'm fairly certain this will come up during review, especially after my conversation with Jin Ma.

As far as I can tell, this matches lui + some constant, but in practice this will usually be a `(high (symbol))`, that's what this pattern is for. I hope that the original `riscv_fuse_lui_ld` *does* cover this case. We should probably reuse this.

The problem with the original `riscv_fuse_lui_ld` is that the constraint on the destination register is different.
Either we outline the checkers in `riscv_fuse_lui_ld` and reuse them. Or we outline the entire `riscv_fuse_lui_ld` and add a paramter to invert the register check.

In any case we should reuse the `riscv_fusion_same_dest_p` util.

gcc/config/riscv/riscv-fusion.cc
```diff
@@ -1174,13 +1177,18 @@ riscv_fuse_lui_ld_rev (rtx_insn *prev, rtx_insn *curr)
   if (!prev_set || !curr_set || any_condjump_p (curr))
     return false;
 
+  /* ARC-V RHX-100 Fused_LD_LUI wants different destinations.  */
+  if (TARGET_ARCV_RHX100
```
luismgsilva: @MichielDerhaeg Does it make sense to add the same/diff destination registers check on `RISCV_FUSE_LUI_LD_REV` ?

I know its only used by RHX-100, but the intent is to keep aligned to `RISCV_FUSE_LUI_LD`

luismgsilva: I tried both approaches on the existing fusion-lui-ld.c cases.

For RISCV_FUSE_LUI_LD, flipping the same/diff-dest check only affects register allocation; the lui+ld shape is unchanged.

For RISCV_FUSE_LUI_LD_REV, using the same riscv_fusion_same_dest_p polarity as LUI_LD changes instruction ordering and can separate the ld/lui pair. Keeping the existing `REGNO (...) != REGNO (...)` check preserves better adjacency.

MichielDerhaeg (approved):

PR #300, opened 2026-07-19
arc64: update mov tests updated gcc16 code-gen

Alexehv77:
arc64: update mov tests updated gcc16 code-gen

gcc16 promotes modes on the callee side during the expand
phase using TARGET_PROMOTE_FUNCTION_MODE. This eliminates
sign extensions on arguments. Already at the expand
level gcc16 produces just a move while gcc15 was making use
of sign_extend of subreg of the same input.
For example sexh_s and extb_s are eliminated in favour of mov_s (see the moveh-2.c). 
Due to those changes already at the expand level gcc16 produces just a move while 
 gcc15 was making use of sign_extend of subreg of the same input.
Because the input parameters arrive for consumtion at the callee side without 
subreg constructs this feature is further positively impacting consumer 
instructions e.g., mpyuw is replaced in gcc16 by mpy (see the case of mpyuw.c).

### DIFF
--- a/gcc/testsuite/gcc.target/arc64/movh-2.c
+++ b/gcc/testsuite/gcc.target/arc64/movh-2.c
@@ -8,4 +8,4 @@ short foo(short a, short b)
 {
   return b;
 }
-/* { dg-final { scan-assembler "sexh_s\\s+r\[0-9\]+,r\[0-9\]+" } } */
+/* { dg-final { scan-assembler "(?:mov_s|sexh_s)\\s+r\[0-9\]+,r\[0-9\]+" } } */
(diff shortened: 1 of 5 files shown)


gcc/testsuite/gcc.target/arc64/movh-2.c
```diff
@@ -8,4 +8,4 @@ short foo(short a, short b)
 {
   return b;
 }
-/* { dg-final { scan-assembler "sexh_s\\s+r\[0-9\]+,r\[0-9\]+" } } */
+/* { dg-final { scan-assembler "\(?:mov_s|sexh_s)\s+r\[0-9\]+,r\[0-9\]+" } } */
```
luismgsilva: This scan-assembler pattern is wrong

```suggestion
/* { dg-final { scan-assembler "(?:mov_s|sexh_s)\\s+r\[0-9\]+,r\[0-9\]+" } } */
```

luismgsilva (changes requested):

MichielDerhaeg (approved):

PR #302, opened 2026-07-20
fixup! RISC-V: Add instruction patterns for 32-bit multiply-add and b…

luismgsilva:
…it-extract fusion.

Do not apply mult-add fusion RTL with XThead MAC enabled

The multiply-add fusion insn_and_split patterns and the maddhisi4 expand hook emit separate mul and add insns so the scheduler can recognize RISCV_FUSE_MULT_ADD pairs.  When the XThead MAC extension is enabled, the target already provides th.mula, th.muls, th.mulah, and related insns for the same RTL shapes.  With a tune that enables mult-add fusion, those hooks could be selected instead of the XThead patterns, breaking gcc.target/riscv/xtheadmac-mula-muls.c.


MichielDerhaeg (approved):
I didn't bother fixing this as this won't happen in in production.
But it doesn't hurt.

PR #306, opened 2026-07-24
riscv: Allow Zcmp push/pop when shrink-wrap-separate is inactive

MichielDerhaeg:
This patch isn't merged yet upstream. I pinged the original author about the status.
This enables zcmp instructions for many more cases.

Not sure if this is still the case, but zcmp used to be slower for RHX. So it likely needs to be gated when we are not optimising for size. But for RMX-500 and 700 this will always be faster, or similar in performance.


MichielDerhaeg:
Put more notes on the Jira: P10019563-31483

exur00:
There are indeed significant regressions for RHX, I'm checking them out

exur00:
For rmx500 there are also regressions (nsichneu, picojpeg, sglib, slre, statemate and ud all betwen 1 and 3 % slower) except nsichneu and statemate, this overlaps with the rhx regressions.

PR #307, opened 2026-07-29
riscv github runner use sourceware git://

exur00:
The previously used https:// rate-limits github runners. The git:// endpoint doesn't seem to do so.


exur00:
Kind of silly work-around, but did succeed for 6 tests in a row, which does seem a lot better than the current situation.

luismgsilva (approved):
LGTM. Lets also apply this patch on `trunk-ci` branch.

PR #311, opened 2026-08-04
arcv: CRC32 for RV32 using clmulr

exur00:
As shown in the code snippet in the RISC-V bitmanip spec v0.93 in section 2.8. It is not present in the ratified spec, but the clmulr definition has not changed between them.

This significantly improves performance for rv32 processors that have the zbc extension.

__builtin_rev_crc32_data32 (crc, data, poly) on rv32_zbc now generates:
	li	a4,-150925312
	xor	a0,a0,a1
	addi	a4,a4,1601
	li	a5,-306675712
	clmul	a0,a0,a4
	addi	a5,a5,800
	clmulr	a0,a0,a5

which is more compact and faster than the previous approach which generates more than 300 lines of assembly and table data.

### DIFF
--- a/gcc/testsuite/gcc.target/riscv/crc-builtin-zbc32.c
+++ b/gcc/testsuite/gcc.target/riscv/crc-builtin-zbc32.c
@@ -19,7 +19,7 @@ int16_t crc16_data16 ()
 }
 
 int32_t rev_crc32_data8 ()
-]{
+{
   return __builtin_rev_crc32_data8 (0x12345678, 'a', 0x04C11DB7);
 }
 
--- a/gcc/testsuite/gcc.target/riscv/crc-builtin-zbc32.c
+++ b/gcc/testsuite/gcc.target/riscv/crc-builtin-zbc32.c
@@ -18,4 +18,20 @@ int16_t crc16_data16 ()
   return __builtin_crc16_data16 (0x1234, 0x3214, 0x1021);
 }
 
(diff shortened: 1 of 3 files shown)


exur00:
https://lkml.org/lkml/2024/6/21/204 the linux kernel also has this in assembly

performance goes from +- 17 cycles/byte to +- 1 cycle per byte for RHX100

exur00:
I think the same could work for crc64 on rv64, but I am not a crc expert, and I can't immediately find a source about it. So I'll leave it for this, don't want to make any mistakes in checksums.

exur00:
let's wait to merge this for the first jenkins run on the gcc16 arc-2026.09

luismgsilva (approved):

gcc/testsuite/gcc.target/riscv/crc-builtin-zbc32.c
```diff
@@ -18,4 +18,20 @@ int16_t crc16_data16 ()
   return __builtin_crc16_data16 (0x1234, 0x3214, 0x1021);
 }
 
-/* { dg-final { scan-assembler-times "clmul\t" 6 } } */
+int32_t rev_crc32_data8 ()
+]{
```
exur00, line 22: typo

MichielDerhaeg:
Pulled down the change that was pushed upstream.
I also included a small fix that is not on GCC 16 yet. Mainly such that the upstream change applies more cleanly.

PR #322, opened 2026-08-26
riscv: add RMX-700 mtune

exur00:
RMX-700 is similar to the RMX-500, it always uses advanced fusions.


gcc/config/riscv/riscv-cores.def
```diff
@@ -53,6 +53,7 @@ RISCV_TUNE("xiangshan-kunminghu", xiangshan, generic_ooo_tune_info)
 RISCV_TUNE("spacemit-x60", spacemit_x60, spacemit_x60_tune_info)
 RISCV_TUNE("arc-v-rmx-100-series", arcv_rmx100, arcv_rmx100_tune_info)
 RISCV_TUNE("arc-v-rmx-500-series", arcv_rmx500, arcv_rmx500_tune_info)
+RISCV_TUNE("arc-v-rmx-700-series", arcv_rmx500, arcv_rmx500_tune_info)
```
MichielDerhaeg: You specify `arcv_rmx700` here, is that correct?

MichielDerhaeg: `rmx500`*

exur00: `RISCV_TUNE(TUNE_NAME, PIPELINE_MODEL, TUNE_INFO)` that specifies the pipeline model, which rmx500 and rmx700 share.

MichielDerhaeg: Then where is `riscv_microarchitecture` set to `arcv_rmx700`?

exur00: very good question, it looks like it's not connected at all, in that case I don't immediately see a way to do this without duplicating the pipeline description.

gcc/config/riscv/riscv.cc
```diff
@@ -4028,7 +4028,7 @@ riscv_legitimize_move (machine_mode mode, rtx dest, rtx src)
 	     already we can just elide the nop move here and be done.  */
 	  if (need_int_reg_p)
 	    emit_move_insn (dest, gen_lowpart (GET_MODE (dest), int_reg));
-	  else if (!rtx_equal_p (dest, int_reg)) 
+	  else if (!rtx_equal_p (dest, int_reg))
```
MichielDerhaeg: I noticed this change in more than 1 PR. Shouldn't be in here.

MichielDerhaeg (approved):

gcc/config/riscv/riscv.cc
exur00: indeed, I guess autoformatting, I will remove it

PR #325, opened 2026-08-28
arc: Fix dmpywh/qmpyh reduction destinations

luismgsilva:
reduc_plus_scal_v2si and reduc_plus_scal_v4hi expanded to dmpywh/qmpyh with an SImode/HImode destination and constraint "r".  Those instructions write a 64-bit result to an even/odd register pair (and the accumulator).

LRA could therefore assign an odd hard register, which raises an instruction error on HS, or silently clobber dest+1 when the destination happened to be even.  gcc.dg/pr92301.c fails this way at -O3 on hs38.

This was exposed by 309dbcea2ca, which replaced power-of-two shifts on the vector factor with multiply/divide when computing vector loop niters.  That changes the peel/niters compare enough that register allocation more readily picked an odd dmpywh destination for this test.

Expand through a DImode temporary so the destination is a real even/odd pair, then take the lowpart.

### DIFF
--- a/gcc/config/arc/simdext.md
+++ b/gcc/config/arc/simdext.md
@@ -2042,9 +2042,20 @@
   [(set_attr "length" "8")
    (set_attr "type" "multi")])
 
-(define_insn "reduc_plus_scal_v4hi"
-  [(set (match_operand:HI 0 "even_register_operand" "=r")
-	(unspec:HI [(match_operand:V4HI 1 "even_register_operand" "r")]
+(define_expand "reduc_plus_scal_v4hi"
+  [(match_operand:HI 0 "register_operand")
+   (match_operand:V4HI 1 "even_register_operand")]
+  "TARGET_PLUS_QMACW"
+  {
+    rtx tmp = gen_reg_rtx (DImode);
+    emit_insn (gen_qmpyh_1 (tmp, operands[1]));
+    emit_move_insn (operands[0], gen_lowpart (HImode, tmp));
+    DONE;
+  })
+
+(define_insn "qmpyh_1"
+  [(set (match_operand:DI 0 "even_register_operand" "=r")
+	(unspec:DI [(match_operand:V4HI 1 "even_register_operand" "r")]
 		   UNSPEC_ARC_QMPYH))
    (clobber (reg:DI ARCV2_ACC))]
   "TARGET_PLUS_QMACW"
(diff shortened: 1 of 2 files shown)


gcc/config/arc/simdext.md
```diff
@@ -2042,19 +2042,41 @@
...
+   (match_operand:V4HI 1 "register_operand")]
+  "TARGET_PLUS_QMACW"
+  {
+    rtx tmp = gen_reg_rtx (DImode);
+    emit_insn (gen_reduc_plus_scal_v4hi_di (tmp, operands[1]));
+    emit_move_insn (operands[0], gen_lowpart (HImode, tmp));
+    DONE;
+  })
+
+(define_insn "reduc_plus_scal_v4hi_di"
```
MichielDerhaeg, line 2056: nit; maybe let's call the insn qpmyh, after the instruction it represents

gcc/config/arc/simdext.md
```diff
@@ -2042,19 +2042,41 @@
...
   [(set_attr "length" "4")
    (set_attr "type" "multi")])
 
-(define_insn "reduc_plus_scal_v2si"
-  [(set (match_operand:SI 0 "even_register_operand" "=r")
-	(unspec:SI [(match_operand:V2SI 1 "even_register_operand" "r")]
+(define_expand "reduc_plus_scal_v2si"
+  [(match_operand:SI 0 "register_operand")
+   (match_operand:V2SI 1 "register_operand")]
+  "TARGET_PLUS_QMACW"
```
MichielDerhaeg, line 2069: The expand here gates on `TARGET_PLUS_QMACW` while the insn is gated by `TARGET_PLUS_DMPY`.
Which should it be?

gcc/config/arc/simdext.md
```diff
@@ -2042,19 +2042,41 @@
   [(set_attr "length" "8")
    (set_attr "type" "multi")])
 
-(define_insn "reduc_plus_scal_v4hi"
-  [(set (match_operand:HI 0 "even_register_operand" "=r")
-	(unspec:HI [(match_operand:V4HI 1 "even_register_operand" "r")]
+(define_expand "reduc_plus_scal_v4hi"
+  [(match_operand:HI 0 "register_operand")
```
MichielDerhaeg, line 2046: Does this also need `even_register_operand`?

MichielDerhaeg (changes requested):
Try to add a test that reproduces the both the "wrong register" and potential clobber issue.

This will go upstream ASAP. Having an upstream-ready commit message would be nice :)

PR #330, opened 2026-09-03
arcv: Update ARC-V target vendor branding.

luismgsilva:


gcc/config.gcc
```diff
@@ -5143,8 +5143,8 @@ case "${target}" in
 		case ${with_tune} in
 		"")
 			case ${target} in
-			*-snps-*)
-				# Use a default mtune for Synopsys ARC-V.
+			*-gf-*)
+				# Use a default mtune for ARC-V.
 				with_tune=arc-v-rmx-100-series
 				;;
```
MichielDerhaeg: You can match both snps and gf for the time being.

MichielDerhaeg (approved):

kolerov (approved):

PR #336, opened 2026-09-09
arc64: fix inline asm m constraint

MichielDerhaeg:
This fixes linux builds.
See https://github.com/foss-for-mips-arc-processors/toolchain/issues/672#issuecomment-5599128217


luismgsilva (approved):

kolerov (approved):
Now Buildroot images for ARCv3 may be built with this fix.

PR #341, opened 2026-09-17
arc64: Fix sr and srl instruction lengths

amikhalev:
## Summary

- Correct the instruction lengths for the arbitrary-immediate and register
  alternatives of the ARC64 `sr` and `srl` patterns.
- Add an assembly regression test covering both instructions.

The alternatives are ordered `U06S0,S12S0,i,r`. The arbitrary immediate uses
a LIMM and encodes in 8 bytes, while the register form encodes in 4 bytes. The
length attributes had those last two sizes reversed.

This made branch shortening underestimate code containing extended auxiliary
register addresses and could select an out-of-range compact branch.

## Testing

- Stock ARC GCC 15.2 fails the new test with:
  - `operand out of range (132 is not between -128 and 127)`
  - `operand out of range (134 is not between -128 and 127)`
- Built a patched stage-1 compiler from `arc-2026.09`; the new `sr` and `srl`
  cases both assemble successfully.


gcc/testsuite/gcc.target/arc64/aux-reg-length.c
```diff
@@ -0,0 +1,76 @@
...
+    wake ();
+}
+
+void
+trigger_srl (void)
+{
+  if (!initialized_srl)
+    initialize_srl ();
+  handle_srl ();
+}
```
luismgsilva, line 76: Besides checking that there are no assembler errors, I think we might as well verify that the correct instruction is emitted. For example:

```suggestion
}

/* { dg-final { scan-assembler "sr\\s+\[^\n\]*\\\[3329\\\]\[^\n\]*l=8" } } */
/* { dg-final { scan-assembler "srl\\s+\[^\n\]*\\\[3329\\\]\[^\n\]*l=8" } } */
```
Of course, this means adding `-dp -save-temps` to the `dg-options`.

luismgsilva (changes requested):
Please add your own sign-off to the commit message.

PR #342, opened 2026-09-21
arcv: recover RHX-100 Dhrystone (sum-of-two-s12 + C2Y typeless revert)

luismgsilva:


MichielDerhaeg:
For reference https://gcc.gnu.org/cgit/gcc/commit/?id=304d7359a67b8f7b979bf5941c7a5cbabf1aebcf https://gcc.gnu.org/bugzilla/show_bug.cgi?id=120811

gcc/config/riscv/riscv.md
```diff
@@ -806,6 +806,53 @@
...
+;; ld/sd with a simm12.  Allowing this pattern on SP/FP lets late-combine
+;; squash that pair back into a single out-of-range PLUS.
+(define_insn_and_split "*add<mode>3_const_sum_of_two_s12"
+  [(set (match_operand:P	 0 "register_operand" "=r,r")
+	(plus:P (match_operand:P 1 "register_operand" " r,r")
+		(match_operand:P 2 "const_two_s12"    " MiG,r")))]
+  "!riscv_reg_frame_related (operands[0])
+   && !riscv_reg_frame_related (operands[1])
+   && !(REG_P (operands[1]) && REGNO (operands[1]) == STACK_POINTER_REGNUM)
+   && !(REG_P (operands[1]) && REGNO (operands[1]) == HARD_FRAME_POINTER_REGNUM)"
```
MichielDerhaeg: I think `riscv_reg_frame_related` implies the next 2 conditions.

MichielDerhaeg (changes requested):
Let's just do the sum-of-two-s12 fix.
I don't think we can carry around the C2Y change anymore and we should just accept the performance hit.

MichielDerhaeg:
For next week. The synthesize thing could use a minimal reproducer so we can report it upstream.

gcc/config/riscv/riscv.md
luismgsilva:
```
/* Helper to determine if reg X pertains to stack.  */
bool
riscv_reg_frame_related (rtx x)
{
  return REG_P (x)
	 && (REGNO (x) == FRAME_POINTER_REGNUM
	     || REGNO (x) == HARD_FRAME_POINTER_REGNUM
	     || REGNO (x) == ARG_POINTER_REGNUM
	     || REGNO (x) == VIRTUAL_STACK_VARS_REGNUM);
}
```
Indeed `riscv_reg_frame_related` it covers HFP, but not  `STACK_POINTER_REGNUM`. I dropped the HFP.

MichielDerhaeg (approved):

